PR #20, opened 2026-04-03
feat: per-component OTA update UI with flash capability detection

justinlindh:
### Pi side
- Detects SWD flash capability at startup (checks for openocd + flash-pico.sh)
- Reports `flash_capable` bool in `device_info` messages to the server

### Server side  
- Stores and exposes `flash_capable` in device info snapshots
- Reworked phone detail page with separate Pi Software and Firmware update sections
- Pi: always shows version picker and update button
- Firmware: shows update button only when device has SWD capability
- Devices without SWD show "Firmware updates require SWD wiring" message
- Green "SWD" / gray "No SWD" badge on firmware label

### Backward compatible
Old Pi binaries that don't send `flash_capable` default to false (no SWD), which is the safe/correct behavior.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -809,7 +809,11 @@ func main() {
 	phone.RestoreVolume()
 	log.Println("digitsd ready")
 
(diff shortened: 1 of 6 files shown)


justinlindh:
@claude simplify

PR #45, opened 2026-04-07
fix(pi): run digitsd from read-only rootfs to prevent corruption

justinlindh:
## What

Moves digitsd to run exclusively from `/usr/local/bin/digitsd` on the read-only rootfs instead of a copy on the writable `/data` partition.

## Why

The digitsd binary on `/data/digits/digitsd/digitsd` was deleted or corrupted on digits-1c09, leaving the service crash-looping (exit 203/EXEC) with no recovery path. The read-only rootfs is already designed to prevent this class of failure -- the binary should live there, not on the writable partition.

Same principle as #44 (wifi config corruption), but simpler: the ro filesystem is the protection, no boot-time validation needed.

## Test plan

- `go vet ./internal/updater/` and `go test ./internal/updater/` pass
- Verified no remaining references to `/data/digits/digitsd` in the codebase
- OTA updater correctly remounts rw, copies the binary (cross-filesystem), and restores ro

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help server server-test pi-build pi-test firmware image image-dev clean
+.PHONY: help server server-test pi-build pi-test firmware image image-dev flash clean
 
 help: ## Show this help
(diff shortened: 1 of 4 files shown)


justinlindh:
@claude

justinlindh:
@claude

PR #47, opened 2026-04-07
fix(digitsd): fix SWD detection path and updater remount permissions

justinlindh:
## What

Fixes two bugs in the web app OTA updater: incorrect SWD capability detection and permission failure when remounting the root filesystem.

## Why

- SWD detection checked `/usr/local/bin/openocd`, but the image installs openocd via apt to `/usr/bin/openocd`. This caused all devices to show "Firmware updates require SWD wiring" even when wiring is present.
- `ApplyPiUpdate()` ran `mount -o remount,rw /` directly, but digitsd runs as `User=digits` (unprivileged), so the mount failed with exit status 32 (EPERM).

## Test plan

- [x] Verified on phone 2 (`192.168.2.228`): openocd exists at `/usr/bin/openocd`, not `/usr/local/bin/openocd`
- [x] Verified `sudo -u digits sudo -n mount -o remount,rw /` succeeds with the new sudoers entry
- [x] Deployed sudoers file to both phones manually for immediate fix
- [x] `make test` passes in `pi/digitsd/`
- [x] Confirmed image builder (`rsync -a`) preserves sudoers file permissions (0440)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,6 +46,20 @@ PICO_SDK_PATH=/path/to/pico-sdk ./scripts/build.sh
 ./scripts/flash.sh  # copies UF2 to mounted Pico
 ```
 
+## Production deployment
+
+The server runs on the GPU box via Docker Compose. All commands from `server/`:
+
+```
+docker compose -f docker-compose.prod.yml --env-file .env.prod up -d      # start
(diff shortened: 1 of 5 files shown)


justinlindh:
@claude

PR #65, opened 2026-04-08
fix(server,digitsd,firmware): minor cleanups across all components

justinlindh:
## What

Small cleanups touching all three components to trigger release CI for each after the tag reset.

## Why

Old release tags pointed to orphaned commits after a history rewrite. Tags were deleted and reset to v1.0.0 on current main. This PR exercises the release pipelines to verify they work with the new tags.

## Test plan

- server: `make test` and `go vet` pass
- digitsd: `go vet` passes
- firmware: change is a trivially correct simplification (redundant ternary where both branches are identical)
- All three release workflows should trigger on merge and produce v1.0.1 releases


justinlindh:
@claude

PR #69, opened 2026-04-08
feat(server): show last seen timestamp for offline devices

justinlindh:
## What

Persists device last-seen timestamps to the database and displays them on the phone detail page for both online ("Last Check-in") and offline ("Last seen") devices.

## Why

Previously, "Last Check-in" only showed when a device was online (sourced from in-memory hub state). When the device went offline or the server restarted, the timestamp was lost. This makes it impossible to tell when an offline device was last connected.

## Test plan

- [x] `go vet ./...` clean
- [x] `make test` passes
- [ ] Manual: connect a device, verify "Last Check-in" shows in Hardware & Software
- [ ] Manual: disconnect device, verify "Last seen" appears with disconnect time
- [ ] Manual: restart server, verify "Last seen" persists for offline devices
- [ ] Manual: verify timestamps display in household timezone

### Changes

- v9 migration: `last_seen_at TIMESTAMPTZ` on devices table (nullable)
- `TouchLastSeen(hardwareID)` store method
- WebSocket handler calls TouchLastSeen on connect, every pong (~30s), and disconnect
- Removed in-memory `LastSeen` from hub `Conn` and `DeviceInfoSnapshot` (DB is source of truth)
- Template shows "Last Check-in" (online) and "Last seen" (offline) from persisted field

### DIFF
--- a/server/internal/db/db.go
+++ b/server/internal/db/db.go
@@ -201,6 +201,8 @@ func (d *Database) migrate() error {
 		`ALTER TABLE devices ALTER COLUMN line_id DROP NOT NULL`,
 		// v8: household timezone
 		`ALTER TABLE households ADD COLUMN IF NOT EXISTS timezone TEXT NOT NULL DEFAULT 'UTC'`,
+		// v9: device last-seen timestamp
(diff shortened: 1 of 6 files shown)


justinlindh:
@claude

PR #71, opened 2026-04-08
fix(server): fix broken JOIN in CanCall cross-household query

justinlindh:
## What

Fix SQL syntax error in the `CanCall` authorizer that broke cross-household call authorization.

## Why

`FROM caller, callee JOIN household_links` is parsed by Postgres as `FROM caller, (callee JOIN household_links)`, making the `caller` CTE inaccessible in the ON clause. Every cross-household authorization check errored, silently falling through to allow-by-default in the relay.

## Test plan

- [ ] Verify linked-household calls succeed without Postgres errors in the DB logs
- [ ] Verify unlinked-household calls are rejected with `not_authorized`


justinlindh:
@claude

justinlindh:
Addressed in another PR, Claude

PR #73, opened 2026-04-08
fix(server): deny calls when CanCall returns an error

justinlindh:
## What

Change the call authorization check in the relay to deny on error instead of allowing.

## Why

The previous check (`err == nil && !allowed`) silently allowed calls through when `CanCall` returned a database error. This is a security issue: any DB failure in the authorization path would bypass call restrictions. Flagged in code review on #71.

## Test plan

- [x] New test `TestRelayCallDeniedOnAuthorizerError` verifies calls are denied when the authorizer returns an error
- [x] All existing relay tests pass
- [x] Full `make test` passes

### DIFF
--- a/server/internal/signaling/relay.go
+++ b/server/internal/signaling/relay.go
@@ -88,6 +88,9 @@ func (r *Relay) handleCall(from string, msg *Message) {
 	if r.CallAuthorizer != nil {
 		allowed, err := r.CallAuthorizer.CanCall(from, msg.To)
 		if err != nil || !allowed {
+			if err != nil {
+				slog.Error("call authorization failed", "from", from, "to", msg.To, "err", err)
+			}
 			r.Hub.SendTo(from, &Message{Type: TypeError, Error: "not_authorized"})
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude review this, and related code for auth/security concerns and flag anything. Before you do so, ensure that you fully understand the related code.

justinlindh:
Synced with main and addressed item 1 (error logging). `slog.Error` now fires when `CanCall` returns an error so DB outages are visible to operators instead of silently denying all calls.

Items 2 and 3 are acknowledged as design observations -- the security boundary is network access, not per-device WS auth, which is acceptable for our threat model. Worth revisiting if the server ever becomes internet-facing.

Item 4 verified correct, no changes needed.

justinlindh:
Correction on item 2: the /ws endpoint is already internet-facing through Caddy with no auth, so this is a current concern, not a hypothetical future one. Opening a follow-up to add device-level authentication on the WebSocket connection immediately.

PR #81, opened 2026-04-09
fix(digitsd): route unsolicited UART events to prevent version query race

justinlindh:
## What

Fixes the serial readLoop to classify unsolicited Pico messages (HOOK, KEY, DIAL, FSM, STATUS:READY) and always route them to the events channel, even when a synchronous command is pending. Also removes the redundant `onFWFlashed` callback and adds a guard against concurrent update runs.

## Why

After a firmware flash on Digits 2, the Pico's `STATUS:READY` boot message was consumed by a pending `VERSION` query's response channel. This caused the version re-query to fail with `unexpected version response: "STATUS:READY"`, leaving `fwVersion` stale. The startup update goroutine (10s delay) then saw firmware as still outdated, triggering a second flash in a loop. The web UI never showed the updated version.

Three issues fixed:
1. **Serial routing**: `readLoop` blindly delivered the next line to `respCh` regardless of message type. Now unsolicited events bypass `respCh` entirely.
2. **Redundant re-query path**: `onFWFlashed` and the `STATUS:READY` event handler both called `QueryVersion()` after a flash, racing on the serial mutex. Removed `onFWFlashed` -- the `STATUS:READY` handler is the single re-query path.
3. **Concurrent updates**: The startup update goroutine could fire mid-flash with a stale version. An `atomic.Bool` guard now prevents overlapping update runs.

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./...` all passing
- [x] New `TestIsUnsolicitedEvent` covers all UART message types from the protocol spec
- [ ] Deploy to Digits 2, trigger firmware flash from web UI, verify version updates without double-flash

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -568,11 +568,21 @@ const (
 // statusFunc is a callback to report update progress back to the server.
 type statusFunc func(status, detail string)
 
+// updateInProgress guards against concurrent firmware update runs.
+// The startup update goroutine and server-triggered updates both check this
+// to avoid racing (e.g. double-flashing the Pico).
+var updateInProgress atomic.Bool
+
 func runUpdate(serverURL, piVersion, fwVersion string, flashCapable bool, reportStatus statusFunc) {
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

PR #84, opened 2026-04-09
test(server): add auth handler and middleware tests

justinlindh:
## What

Add unit tests for the previously untested auth handlers, middleware, and Google OAuth code in `server/internal/auth/`.

## Why

The auth package had test coverage only for `store.go`. The HTTP handlers (`handlers.go`), middleware (`middleware.go`), and Google OAuth (`google.go`) had zero tests, leaving critical authentication logic unverified.

## Test plan

- `cd server && make test` passes
- `cd server && go vet ./...` clean
- New tests cover:
  - **middleware**: valid session (user injected into context), invalid/expired session (redirect + cookie cleared), missing cookie (redirect)
  - **handlers**: login page rendering with error/success params, magic link request (empty email rejected, valid email sends mail), magic link verify (invalid token rejected, valid token creates user + session), logout (clears session + cookie)
  - **google**: `Enabled()` with various credential combos, `HandleLogin` sets state cookie + redirects, `HandleCallback` rejects missing/mismatched state

### DIFF
--- a/server/internal/auth/google_test.go
+++ b/server/internal/auth/google_test.go
@@ -6,30 +6,6 @@ import (
 	"testing"
 )
 
-func TestGoogleAuth_Enabled(t *testing.T) {
-	s := testDB(t)
-
-	tests := []struct {
-		name     string
-		clientID string
-		secret   string
-		want     bool
-	}{
-		{"both set", "id", "secret", true},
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude - review these; are they all worth having, or are some low/zero value?

PR #86, opened 2026-04-09
fix(server): handle swallowed errors in web handlers

justinlindh:
## What

Add proper error handling for operations in `server/internal/web/handler.go` that previously silently ignored errors.

## Why

Many web handlers swallowed errors from `ParseForm()`, `lineStore.Delete()`, `householdStore.UpdateName()`, `SetCallHistoryEnabled()`, `SetTimezone()`, stats endpoint queries, and `json.Encode()` calls. Users would see success responses even when the underlying operations failed, making debugging difficult and hiding real problems.

## Test plan

- `make test` passes
- `go vet ./...` clean
- Reviewed each changed call site to confirm error handling follows the existing pattern for that handler (400 for parse errors, 500 for store errors, redirect without saved flag for settings errors, log for json encode errors on already-written responses)

### DIFF
--- a/server/internal/line/store.go
+++ b/server/internal/line/store.go
@@ -2,12 +2,16 @@ package line
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"regexp"
 	"time"
 
 	"github.com/justinlindh/digits/server/internal/db"
 )
+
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

justinlindh:
@claude - happy now?

PR #88, opened 2026-04-09
fix(digitsd): eliminate UART event race and concurrent update guard

justinlindh:
## What

Fix a race condition where the Pico's `STATUS:READY` boot message was consumed by a pending `VERSION` query's response channel after firmware flash, causing the version re-query to fail and triggering a flash loop.

## Why

After a firmware flash, `STATUS:READY` is an unsolicited event from the Pico, but the serial layer was routing it to whichever synchronous command was waiting for a response. This left `fwVersion` stale, which made the startup update goroutine think the firmware was outdated and flash again in a loop.

## Changes

- **`serial.go`**: Add `isUnsolicitedEvent()` classifier so known Pico events (`HOOK`, `KEY`, `DIAL`, `FSM`, `STATUS:READY`) always route to the events channel, even when a synchronous command is pending
- **`serial.go`**: Switch `respCh` from mutex-guarded pointer to `atomic.Pointer[chan string]` to eliminate a data race between `SendCommand` and `readLoop`
- **`main.go`**: Add `updateInProgress` atomic guard to prevent concurrent update runs
- **`main.go`**: Remove `onFWFlashed` callback (now redundant with the `STATUS:READY` event handler)

## Test plan

- `make test` passes
- `go vet ./...` clean
- Verified on device: firmware flash completes, `STATUS:READY` triggers version re-query, no flash loop

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -568,11 +568,21 @@ const (
 // statusFunc is a callback to report update progress back to the server.
 type statusFunc func(status, detail string)
 
+// updateInProgress guards against concurrent firmware update runs.
+// The startup update goroutine and server-triggered updates both check this
+// to avoid racing (e.g. double-flashing the Pico).
+var updateInProgress atomic.Bool
+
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

PR #91, opened 2026-04-09
fix(digitsd): match POTS Bellcore timings for remote hangup

justinlindh:
## What

Corrects the remote hangup off-hook sequence to match real US POTS behavior per Bellcore GR-506-CORE permanent signal treatment.

## Why

The previous timings didn't match what a real phone line does when the far end hangs up:
- Reorder tone was 10s instead of ~45s
- An intercept message (SIT tones) played between reorder and howler -- intercept is for invalid/disconnected numbers, not far-end hangup
- Howler ran indefinitely with no line lockout

Corrected sequence: 1s silence, 45s reorder, 3 min howler, then line lockout (silence until hangup).

## Test plan

- Existing unit tests pass (`go test ./internal/phone/...`)
- Manual test: place a call between two phones, hang up one end, verify the remaining handset hears ~45s of reorder tone followed by howler, then silence after ~3 min

### DIFF
--- a/pi/digitsd/internal/phone/controller.go
+++ b/pi/digitsd/internal/phone/controller.go
@@ -297,15 +297,15 @@ func (c *Controller) onSignalHangup() {
 		log.Printf("phone: hangup signal ignored in state %s (not CONNECTED)", c.state)
 		return
 	}
-	// Full POTS remote-hangup sequence:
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

PR #94, opened 2026-04-09
refactor(digitsd): migrate from log to log/slog

justinlindh:
## Summary
- Migrate all 17 Go files in `pi/digitsd/` from the stdlib `log` package to `log/slog`
- `log.Printf` / `log.Println` mapped to `slog.Info`, `slog.Debug`, or `slog.Error` based on context
- `log.Fatalf` / `log.Fatal` replaced with `slog.Error` + `os.Exit(1)`
- No structured key-value fields added -- messages kept as plain strings with `fmt.Sprintf` where needed

## Test plan
- [x] `go vet ./...` passes
- [x] `make test` passes


justinlindh:
@claude

justinlindh:
Recreating with clean branch to fix commitlint (no merge commit).

PR #98, opened 2026-04-09
feat(pi): add deploy-phone skill

justinlindh:
## What

Adds a Claude Code skill for deploying built binaries to phone devices over SSH. Covers all deployable Pi binaries with their build quirks, the read-only rootfs remount cycle, and post-deploy verification.

## Why

Deploying to phones requires a specific sequence (scp to /tmp, remount rw, mv, chmod, remount ro, restart service) that varies by binary. Each binary has different build commands, output names, and install destinations. This skill gives the agent a complete reference so it gets the details right without trial and error.

## Test plan

- Tested with subagent dry runs across three scenarios: digitsd to both phones, digits-setup to one phone, diagnostic tool to /tmp
- Verified skill correctly guides build command selection (make build vs make build-pi vs direct go build)
- Verified digits-setup arm64 naming quirk is handled
- Verified diagnostic tools deploy to /tmp without remount cycle

### DIFF
--- a/.claude/skills/deploy-phone/SKILL.md
+++ b/.claude/skills/deploy-phone/SKILL.md
@@ -1,0 +1,73 @@
+---
+name: deploy-phone
+description: Use when deploying built binaries to phone devices over SSH, updating digitsd or other Pi binaries on one or more phones, or when the user says "deploy" in the context of phone hardware
+---
+
+# Deploy to Phone
+
+Build and deploy Pi binaries to phone devices over SSH. Handles the read-only rootfs remount cycle.
+
+## Deployable Binaries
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude - think this would help you for local dev?

justinlindh:
Good catch on the divergence. The Makefile `deploy` target is an older dev workflow that copies to `~/digits/digitsd` for manual testing. The skill targets `/usr/local/bin/digitsd`, which is where the systemd service runs from on the production image. They serve different purposes -- the skill is for "install as system binary and restart the service," not "drop it somewhere to run by hand."

PR #101, opened 2026-04-09
feat: remote device restart from web UI

justinlindh:
## What

Adds controls to the phone detail page to remotely restart the digitsd service or reboot the entire Pi device. Also removes the automatic update check on startup so updates are always user-initiated.

## Why

No way to restart a device without SSH access. The auto-updater was also silently overwriting dev/test binaries with released versions, making testing impossible.

## Test plan

- [x] Tested "Restart Service" on Digits 1 from app.digits.family -- service restarted, device reconnected, UI showed success
- [x] Tested "Reboot Device" on Digits 1 -- full reboot completed, device came back online
- [x] Confirmed inline confirmation flow works (both buttons hide, full-width confirm replaces them)
- [x] Confirmed error state shows message with "Try again" link
- [x] Confirmed auto-updater no longer runs on startup
- [x] `go test ./...` and `go vet ./...` pass for both server and digitsd
- [ ] Verify `*#UPDATE#` service code still triggers updates (preserved path)
- [ ] Verify web UI "Update" buttons still work (preserved path)

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -567,10 +567,6 @@ type statusFunc func(status, detail string)
 // to avoid racing (e.g. double-flashing the Pico).
 var updateInProgress atomic.Bool
 
-func runUpdate(serverURL, piVersion, fwVersion string, flashCapable bool, reportStatus statusFunc) {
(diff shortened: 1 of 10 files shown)


justinlindh:
@claude

PR #111, opened 2026-04-09
feat: asset embedding and factory reset system

justinlindh:
## What

Makes digitsd the single update vector for all device files by embedding tones, scripts, configs, and systemd units via Go's `embed` package. Adds a full factory reset system with a recovery partition, standalone recovery binary, and server-side trigger.

## Why

Shipped devices have no SSH access and no SD card access. digitsd is the only way to change anything on the device after shipping. This change ensures all deployable files travel with the binary, and provides a recovery mechanism if something goes wrong -- including a factory reset that restores the device to its original state.

## Test plan

- [x] All digitsd tests pass (`go test ./...`) including new tests for assets, watchdog, and bootcount packages
- [x] All server tests pass
- [x] Recovery binary tests pass
- [x] `go vet` clean across all modules
- [x] `make embed` generates 52 files from rootfs-overlay and tones
- [x] Code review caught and fixed: embed.FS path mismatch (asset extraction was silently broken), missing protocol constant, unchecked errors in recovery binary, doc inaccuracies
- [ ] Image build with 4-partition layout (requires Pi hardware)
- [ ] End-to-end: OTA update extracts assets on version change
- [ ] End-to-end: factory reset via web UI triggers recovery mode
- [ ] End-to-end: 3 failed boots enters recovery mode automatically
- [ ] End-to-end: recovery web UI "Try Again" and "Factory Reset" work


justinlindh:
Still showing inflated diff, need to fix branch history first.

PR #117, opened 2026-04-09
fix(digitsd): reconnect to refresh pairing code before expiry

### LINKED ISSUE #116: fix(digitsd): pairing code not refreshed after expiry
## Bug

When a pairing code expires, digitsd does not request or log a new one. The only pairing code in the journal is the one issued at initial connect. If the user doesn't pair before it expires, there's no way to get a fresh code without restarting digitsd.

## Expected

digitsd should either:
- Automatically request a new pairing code when the current one expires, or
- Log the new code when the server pushes one

## Observed

```
Apr 09 21:30:05 digits-f458 digitsd[1156]: INFO PAIRING REQUIRED: pick up handset to hear it code=542034
```

No subsequent pairing code entries appear in the journal. The web UI reports "invalid" when using this code minutes later.

## Repro

1. Flash a fresh image, let device boot and connect to server
2. Wait several minutes without pairing
3. Attempt to pair with the code from the journal -- fails as invalid
4. `journalctl -u digitsd | grep PAIRING` shows only the original code

justinlindh:
Closes #116

When a pairing code expires (server-side 10 min TTL), digitsd had no mechanism to obtain a fresh one. Add a 9-minute refresh timer that triggers a reconnect while the device is still unpaired, causing the server to issue a new code on re-registration.

Generated with [Claude Code](https://claude.ai/code)


justinlindh:
@claude

justinlindh:
@claude - more changes made. Let's review them to ensure they're correct.

PR #120, opened 2026-04-10
fix(pi): resolve AP SSID placeholder before starting hostapd

### LINKED ISSUE #118: fix(pi): AP SSID shows literal 'Digits-XXXX' instead of generated suffix
## Bug

On fresh image flash, the AP mode SSID is `Digits-XXXX` with a literal XXXX instead of a generated suffix (e.g., `Digits-e3b4`). Previously the SSID was set to a unique suffix derived from the device ID.

## Expected

`digits-first-boot.sh` generates a device ID and updates the hostapd SSID to `Digits-<suffix>`. The AP should broadcast with the generated suffix.

## Likely cause

Timing issue: `digits-first-boot.service` has `ConditionPathExists=!/data/.initialized`, but the data skeleton from the image build may already contain `.initialized`, causing first-boot to be skipped. Without first-boot, the hostapd config retains the placeholder SSID.

Alternatively, first-boot may run but the SSID update happens after hostapd has already started.

## Observed on

Fresh flash from `feat/asset-embed-factory-reset` branch image build.

justinlindh:
Closes #118

## Summary

- Add defensive SSID resolution in `digits-ap-check` to replace the build-time placeholder `Digits-XXXX` before starting hostapd
- Add `Wants=digits-first-boot.service` to `digits-ap-check.service` for explicit systemd dependency

## Test plan

- [ ] Flash a fresh image and verify the AP SSID shows a generated suffix (e.g. `Digits-e3b4`) instead of `Digits-XXXX`
- [ ] Verify that subsequent reboots retain the generated SSID
- [ ] Test factory reset flow to confirm SSID is re-generated

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pi/image/rootfs-overlay/etc/systemd/system/digits-ap-check.service
+++ b/pi/image/rootfs-overlay/etc/systemd/system/digits-ap-check.service
@@ -3,6 +3,7 @@ Description=Digits AP Mode Boot Check
 Documentation=https://github.com/digits/digits
 # Run after filesystem and first-boot are done, before networking
 DefaultDependencies=no
+Wants=digits-first-boot.service
 After=local-fs.target digits-first-boot.service
 Before=network-pre.target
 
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #129, opened 2026-04-10
fix(server): move device last-seen tracking to in-memory Hub

justinlindh:
## Summary
- Track device last-seen timestamps in memory on the Hub's `Conn` struct, updated on every WebSocket pong with zero DB cost
- Phone detail page reads the real-time in-memory timestamp for online devices, falls back to the DB `last_seen_at` column for offline devices
- Write `last_seen_at` to Postgres only once on disconnect, so offline devices still show a meaningful timestamp
- Remove all periodic DB writes from the pong handler (previously wrote every ~30s per device)

## Test plan
- [x] `go build ./...` passes
- [x] `go test -race ./...` passes
- [x] `go vet ./...` passes
- [ ] Verify online devices show a recent "last seen" timestamp on phone detail page
- [ ] Verify offline devices show last disconnect time
- [ ] Confirm no `TouchLastSeen` DB writes during normal WebSocket operation (only on disconnect)

### DIFF
--- a/server/internal/signaling/hub.go
+++ b/server/internal/signaling/hub.go
@@ -14,6 +14,7 @@ type Conn struct {
 	Number     string
 	HardwareID string
 	Send       chan []byte
+	LastSeen   time.Time
 
 	// Device info (reported on connect via device_info message)
 	PiVersion       string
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #131, opened 2026-04-10
fix(server): throttle device last-seen updates to reduce DB writes

justinlindh:
## Summary
- Adds in-memory throttle (5 min) to `TouchLastSeen` DB writes in the WebSocket pong handler
- Connect and disconnect `TouchLastSeen` calls remain unthrottled so device status transitions are always recorded
- Reduces unnecessary DB pressure from pong-driven writes (~every 30s per device down to at most every 5 min)

## Test plan
- [x] `go build ./...` passes
- [x] `go test -race ./...` passes
- [x] `go vet ./...` passes
- [ ] Verify in staging that device last_seen timestamps still update within 5 minutes of a device connecting
- [ ] Confirm DB write volume drops for long-lived WebSocket connections

### DIFF
--- a/server/internal/web/handler.go
+++ b/server/internal/web/handler.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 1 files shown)


justinlindh:
Duplicate of #129 (created by a late-arriving retry agent)

PR #132, opened 2026-04-10
fix(server): batch line queries on family page to fix N+1

justinlindh:
## Summary
- Replace N+1 `ListByHousehold` calls inside the linked-households loop on the family page with a single `ListByHouseholds` batch query using `ANY($1)`.
- Add `ListByHouseholds(householdIDs []string)` method to `line.Store` that returns lines grouped by household ID in one query.

## Test plan
- [ ] Verify family/links page renders correctly with linked households
- [ ] Verify page works with zero linked households (empty slice path)
- [ ] Existing tests pass (`go test -race ./...`)

### DIFF
--- a/server/internal/line/store.go
+++ b/server/internal/line/store.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/lib/pq"
+
 	"github.com/justinlindh/digits/server/internal/db"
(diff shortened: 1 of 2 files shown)


justinlindh:
Duplicate of #130 (created by a late-arriving retry agent)

PR #151, opened 2026-04-12
fix(hardware): add LCSC part numbers for JST ZH connectors

justinlindh:
## Summary
- Add LCSC numbers for J4 (C265294), J6 (C265284), J8 (C265083) in schematic
- Enables JLCPCB to place the SMT JST ZH connectors during assembly (already done on V1 boards)

## Test plan
- [ ] Update PCB from schematic and regenerate BOM before next order


justinlindh:
Replaced with clean branch

PR #155, opened 2026-04-12
fix(digitsd): retry Pico VERSION query after reboot

justinlindh:
## Summary

After flashing the Pico over SWD, `digitsd` sees `STATUS:READY` and immediately queries `VERSION`, but the firmware's UART command loop is not yet accepting commands that early in boot. The query times out, the warning `pico: version query after reboot failed error="unexpected version response: \"\""` is logged, and the cached `fw_version` stays at the pre-flash value. The web UI then keeps showing the old firmware version even though the flash succeeded.

Retry the query up to 5 times with 500 ms backoff. A subsequent query outside this window reliably succeeds in practice.

Also makes `SerialPort.SendCommand` return an explicit timeout error instead of `("", nil)` on timeout, which was the source of the confusing `"unexpected version response: \"\""` log line (the caller couldn't distinguish "no reply" from "malformed reply").

Caught in the field on Digits 1 this afternoon: firmware was flashed to 1.2.0 twice via the web-UI update trigger, both flashes actually succeeded (verified by querying `VERSION` over the UART socket directly), but the daemon kept reporting 1.1.0.

## Test plan

- [x] `go build ./...` in `pi/digitsd/`
- [x] `go test ./...` in `pi/digitsd/`
- [x] `golangci-lint run ./...` in `pi/digitsd/`
- [ ] Deploy to Digits 1, trigger an update that re-flashes firmware, confirm web UI picks up the new version without a manual `systemctl restart digitsd`

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -1187,6 +1187,14 @@ func main() {
 	quit := make(chan os.Signal, 1)
 	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
 
+	// Post-reboot firmware version results. STATUS:READY spawns a goroutine
+	// that retries QueryVersion (the Pico's UART command loop isn't quite
+	// awake yet at the instant it emits STATUS:READY). The goroutine sends
+	// the result here so the main loop can update fwVersion/fwCommit without
+	// any shared-state synchronization.
+	type fwVersionResult struct{ version, commit string }
+	fwVersionCh := make(chan fwVersionResult, 1)
+
 	// Main select loop
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the careful review. Addressed all three points in 824e383:

1. **Blocking the event loop** — the retry is now kicked off in a goroutine that delivers its result via a dedicated `fwVersionCh` handled by a new case in the main `select`. Reads and writes of `fwVersion`/`fwCommit` all stay on the main goroutine, so no mutex needed. HOOK/KEY events continue to drain normally while the retry runs.

2. **Trailing sleep after final attempt** — fixed; `time.Sleep` is now guarded by `if attempt < attempts`.

3. **Log-search conflation** — per-attempt message renamed to `"pico: version query attempt failed"`; the terminal message remains `"pico: version query after reboot gave up"`.

PR #156, opened 2026-04-12
fix(digitsd): content-hash marker for asset extraction

justinlindh:
## Summary

The asset extractor at \`pi/digitsd/internal/assets/assets.go:33\` gates re-extraction on \`version.Version\` alone. If a release is rebuilt or hotfixed under the same version tag, devices whose marker already matches that string skip extraction forever and keep stale files on disk.

Caught in the field on Digits 1 today:

- \`/data/digits/asset-version\` said \`1.8.3\`
- \`/etc/systemd/system/digits-mixer.service\` on disk was the **pre-#143** version (no \`ExecStartPre\` wait loop)
- The \`digitsd\` binary running on the device was also \`1.8.3\`, built from commit \`9110e7b\`, which is **after** PR #143 merged — so its embedded overlay had the fix
- On every boot, the extractor saw marker == version == \`1.8.3\`, logged \`assets: version 1.8.3 matches, skipping extraction\`, and left the broken unit file in place
- The broken unit then raced the \`da7213\` codec probe at boot. Whether audio worked was pure luck (whichever defaults the codec came up with before \`alsactl restore Zero\` failed with "Cannot find soundcard 'Zero'")

Compute a SHA-256 over the sorted embedded FS contents and include it in the marker as \`<version>:<hash>\`. Re-extract whenever the hash changes, regardless of version.

## Regression test

Adds \`TestExtract_ReExtractsWhenContentChangesUnderSameVersion\` which reproduces the exact failure: extracts one FS under \`1.0.0\`, then extracts a different FS under the same \`1.0.0\`, and asserts the on-disk file got updated. Would have caught this bug.

Also renames \`TestExtract_SkipsWhenVersionMatches\` → \`TestExtract_SkipsWhenMarkerMatches\` and updates it to reflect the new marker semantics.

## Test plan

- [x] \`go build ./...\` in \`pi/digitsd/\`
- [x] \`go test ./...\` in \`pi/digitsd/\`
- [x] \`golangci-lint run ./...\` in \`pi/digitsd/\`
- [ ] Deploy to Digits 1, confirm the stale \`digits-mixer.service\` on disk gets overwritten with the current embedded version on first boot after upgrade
- [ ] Confirm \`/data/digits/asset-version\` contains a \`<version>:<hash>\` string afterwards

### DIFF
--- a/pi/digitsd/internal/assets/assets.go
+++ b/pi/digitsd/internal/assets/assets.go
@@ -1,11 +1,14 @@
 package assets
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io/fs"
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #159, opened 2026-04-12
fix(digitsd): harden asset extractor against partial failures

justinlindh:
## Summary

Two robustness improvements to `internal/assets/Extractor`, both driven by failure modes I hit while chasing the audio regression on Digits 1 earlier today:

### 1. Log-and-continue past individual file errors

Before, the extractor returned on the first `writeFileWith` error. One bad file would:

- Skip every later rootfs file
- Skip the entire data/* pass
- Skip the marker write
- Mask the partial success in a single error return

This bit me when testing PR #158: `/data/digits/tones/pairing/*` was owned by `pi:pi` from an older install, digitsd runs as `digits`, so `os.WriteFile` got `EACCES` on the first pairing file and the whole extract bailed. Everything earlier in the walk (including the newly-fixed `digits-mixer.service`) had been written, but the marker wasn't, so the next boot would try again and hit the same wall.

Now: log each failure, continue the walk, return an aggregated error at the end. Marker is still withheld on any failure so the next boot gets another chance to retry — important because persistent failures (like the ownership issue) need a human to unblock them, and we don't want to silently declare success.

### 2. `systemctl daemon-reload` after rootfs writes

Even when extraction succeeds, systemd doesn't notice that unit files under `/etc/systemd/system/` changed until it re-reads them. So a rewritten `digits-mixer.service` would sit on disk without actually taking effect until the next reboot. I hit this during Digits 1 testing — had to manually `daemon-reload` + `start digits-mixer.service` after the new binary wrote the fixed unit.

Plumbed as a new `ReloadSystemd func() error` callback on the `Extractor` struct so the package stays free of `exec` and `sudo` concerns. `cmd/digitsd/main.go` wires it to `exec.Command("sudo", "systemctl", "daemon-reload")`. The callback fires unconditionally whenever any rootfs file was written (not gated on whether the path was under `/etc/systemd/system/`): daemon-reload is idempotent and ~100 ms, so path detection adds complexity without meaningful benefit.

## Tests

Three new tests in `assets_test.go`:

- `TestExtract_ContinuesPastFileErrors`: injects a WriteFunc that fails on exactly one rootfs file, asserts the other rootfs files AND all data files still get written, asserts the error mentions the failing file, asserts the marker was withheld.
- `TestExtract_CallsReloadSystemdAfterRootfsWrites`: asserts the callback fires exactly once when a rootfs file is extracted.
- `TestExtract_SkipsReloadWhenNoRootfsFiles`: asserts the callback does not fire when the embedded FS contains only `data/*`.

Existing tests still pass unmodified.

## Related

- #156 (content-hash marker) — merged in pi/v1.8.4
- #158 (populate embedded assets in release build) — open

This completes the extractor hardening items I flagged as follow-ups in #158.

## Test plan

- [x] `go vet ./...`, `go build ./...`, `go test ./...`, `golangci-lint run ./...` all clean in `pi/digitsd/`
- [ ] Merge and cut a new `pi/v1.8.5` (which should also carry #158 once that lands). OTA to Digits 1 and confirm:
  - extractor runs without manual intervention
  - `digits-mixer.service` gets reloaded and runs to completion during the same boot (not the reboot after)

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -766,6 +766,9 @@ func main() {
 			}
 			return nil
 		},
+		ReloadSystemd: func() error {
+			return exec.Command("sudo", "systemctl", "daemon-reload").Run()
+		},
 	}
 	if err := extractor.Extract(version.Version); err != nil {
 		slog.Warn("asset extraction failed", "err", err)
--- a/pi/digitsd/internal/assets/assets.go
+++ b/pi/digitsd/internal/assets/assets.go
@@ -25,12 +25,16 @@ var permOverrides = []struct {
 type WriteFunc func(data []byte, dest string, perm os.FileMode) error
 
 type Extractor struct {
-	FS             fs.FS
-	RootDir        string
-	DataDir        string
-	MarkerPath     string
-	Remount        func(rw bool) error
+	FS              fs.FS
+	RootDir         string
+	DataDir         string
+	MarkerPath      string
+	Remount         func(rw bool) error
 	RootfsWriteFile WriteFunc // privileged writer for rootfs files (nil = use os.WriteFile)
+	// ReloadSystemd, if set, is called after any rootfs files are written so
+	// that systemd picks up rewritten unit files without requiring a reboot.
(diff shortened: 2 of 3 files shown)


justinlindh:
@claude

PR #161, opened 2026-04-13
fix(digitsd,firmware,server): relay in-call keypad presses and fix FSM + ringback bugs

justinlindh:
## Summary

- Fix a caller-side FSM bug: the Pico firmware never reached `PHONE_STATE_CONNECTED` because nothing told it the call was answered. Adds a `CALL:CONNECTED` UART command sent by digitsd on `onSignalAnswer`.
- Relay keypad presses during a connected call to the remote peer via a new `TypeDTMF` signaling message, so both sides hear the tone (matches real phone behavior). Previously: caller heard their own presses locally, callee heard nothing, and the callee's presses were silently dropped by firmware.
- Fix a ringback cancellation bug: `daemonCallbacks.InitiateCall` called `mixer.StopTone()` microseconds after the controller emitted ringback, killing it. Deleted the stale `StopTone()` call; ringback is stopped at the correct moment by `onSignalAnswer`.

## Test plan

- [x] `go test ./...` passes in both `pi/digitsd` and `server` modules
- [x] `go vet ./...` clean in both modules
- [x] `golangci-lint run ./...` clean in both modules
- [x] Firmware builds cleanly via `./scripts/build.sh` (digits.uf2 produced)
- [ ] Manual end-to-end on Digits 1 <-> Digits 2:
  - Place call Digits 1 -> Digits 2; ringback audible on caller until answered
  - After answer, press digits on Digits 1; tones heard on both handsets
  - Press digits on Digits 2; tones heard on both handsets
  - Hang up from either side; both FSMs return to IDLE cleanly

## Notes

- The callee path intentionally does not send `CALL:CONNECTED`: callee firmware already self-transitions `RINGING -> CONNECTED` on hook-off, so only the caller needs the explicit notification.
- DTMF is relayed peer-to-peer through the existing signaling server relay, not over a WebRTC data channel. The signaling server just routes the new `TypeDTMF` message via its existing `forward` helper.
- Also updates CLAUDE.md to note that `./scripts/build.sh` auto-detects `PICO_SDK_PATH` from common system locations.


justinlindh:
@claude

PR #162, opened 2026-04-13
feat(firmware): add Docker-based firmware build

justinlindh:
## Summary

- Add \`firmware/Dockerfile.builder\` and \`firmware/Makefile\` so the Pico firmware can be cross-compiled without installing \`arm-none-eabi-gcc\` or the Pico SDK on the host. Mirrors the pattern already used for \`pi/digitsd/Dockerfile.builder\`.
- \`make firmware\` (from repo root) now runs the Docker build with no host toolchain required.
- \`make firmware-local\` preserves the existing host build via \`scripts/build.sh\` (which auto-detects \`PICO_SDK_PATH\` at common locations).
- \`docker run\` uses \`--user \$(id -u):\$(id -g)\` so build outputs are owned by the invoking user and \`make clean\` works without sudo.

## Dockerfile contents

- \`debian:bookworm-slim\` base
- \`build-essential gcc-arm-none-eabi libnewlib-arm-none-eabi libstdc++-arm-none-eabi-newlib cmake python3 git\`
- pico-sdk cloned at \`/opt/pico-sdk\` with submodules
- \`PICO_SDK_PATH\` preset

\`build-essential\` is required because picotool (a host tool the Pico SDK builds during firmware compilation) needs a native C/C++ compiler, not just the ARM cross-compiler.

## Test plan

- [x] \`make firmware\` builds \`firmware/build/digits.uf2\` from a clean checkout
- [x] Output files owned by invoking user, not root
- [x] \`make -C firmware clean\` succeeds without sudo
- [x] \`make firmware-local\` still works unchanged
- [ ] Manual flash of resulting UF2 to a Pico for sanity (not required, no code changed)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,9 +42,12 @@ make test
 
 Firmware (from repo root):
 ```
-PICO_SDK_PATH=/path/to/pico-sdk ./scripts/build.sh
-./scripts/flash.sh  # copies UF2 to mounted Pico
+make firmware         # builds in Docker, no host toolchain needed
+make firmware-local   # builds on host (requires arm-none-eabi-gcc + Pico SDK)
+./scripts/flash.sh    # copies UF2 to mounted Pico
 ```
+
+`make firmware-local` runs `./scripts/build.sh`, which auto-detects `PICO_SDK_PATH` from common locations (`/usr/share/pico-sdk`, `~/pico-sdk`, etc.). Only set `PICO_SDK_PATH` explicitly if the SDK is installed somewhere non-standard.
 
 ## Linting
 
--- a/Makefile
+++ b/Makefile
@@ -65,4 +65,5 @@ test: server-test pi-test ## Run all tests
 clean: ## Clean build artifacts
 	$(MAKE) -C server clean
 	$(MAKE) -C pi/digitsd clean
-	rm -rf tools/build/ firmware/build/
+	$(MAKE) -C firmware clean
+	rm -rf tools/build/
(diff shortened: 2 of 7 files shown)


justinlindh:
@claude

justinlindh:
Addressed all three blocking issues plus the build-essential comment nit in ac4e37b.

**1. Stale CMake cache across modes** — split into `firmware/build/docker` and `firmware/build/local`. Verified round-trip: `make firmware` → `make firmware-local` → `make firmware` works with no cache poisoning; both UF2s coexist. `scripts/build.sh` now passes the firmware dir to `cmake` explicitly so the deeper subdir works. `scripts/flash.sh` picks whichever UF2 is newer (or whichever exists). `scripts/deploy.sh` targets `build/local` since it always runs `build.sh` first.

**2. pico-sdk unpinned** — pinned to `2.2.0` via `ARG PICO_SDK_VERSION=2.2.0` in the Dockerfile (matches the version installed on the dev host). The build arg means future bumps don't need a Dockerfile edit.

**3. Root Makefile clean bypasses firmware Makefile** — now delegates to `$(MAKE) -C firmware clean`.

**Nit (build-essential comment)** — added: `# build-essential is required: picotool is built during firmware compilation and needs a native (not cross) C/C++ compiler.`

Passed on the two non-blocking nits (dropping `-q` from `docker build`, handling non-git tarball checkouts) to stay consistent with the existing `pi/digitsd/Dockerfile.builder` pattern.

Tested end-to-end:
- Fresh `make firmware` → `firmware/build/docker/digits.uf2` (80KB, owned by user)
- `make firmware-local` afterward → `firmware/build/local/digits.uf2` (82KB, owned by user) with no cache collision
- `make firmware` again after local build → still works, no collision
- `make -C firmware clean` nukes both subdirs

PR #165, opened 2026-04-13
feat(digitsd): add audio filter library and audiotest wav dump

justinlindh:
## Summary

- Adds an IIR biquad filter library (HPF, LPF, notch via RBJ cookbook) with a `BiquadChain` cascade and a `NewPOTSChain` factory that builds a 300-3400 Hz telephony bandpass + 60 Hz harmonic notch comb. Wired behind a new `PipelineConfig.Bandpass` flag, defaulted **off** — RNNoise alone handles the prototype phones' mains hum and preamp hiss more cleanly than any fixed-coefficient filter in front of it. The library stays in the tree as an opt-in tool for future POTS voice-character shaping or codecs without RNNoise.
- `*#8378#` (`*#TEST#`) now dumps the recorded buffer to `/tmp/audiotest_last.wav` so samples can be pulled via `scp` for offline FFT / filter design / subjective A/B without re-recording each time.
- Self-test now runs the default pipeline config (Denoise on) instead of forcing Denoise off, so `*#TEST#` playback mirrors what the remote peer hears in a real call.

Investigation notes that led here: raw-mic FFT showed 60 Hz mains hum at -48 dBFS with odd-harmonic structure (signature of magnetic coupling into the unshielded TRS run, which the existing 100 nF bypass cap cannot touch at 60 Hz). Initial attempts to fix it with biquad bandpass + notch comb in front of RNNoise introduced audible voice artifacts (comb eating speech harmonics whose pitch divided 60 evenly, plus broadband noise floor still leaking through). Running RNNoise alone pulls 60 Hz to -70 dBFS and the broader voice-band floor below audibility, with no perceptible speech distortion.

## Test plan

- [x] `go test ./...` green in `pi/digitsd/`
- [x] `go vet ./...` clean
- [x] Built via `make build` (Docker cross-compile)
- [x] Deployed to both prototype phones, verified rootfs `ro` post-deploy
- [x] `*#8378#` on digits 2: subjectively flawless (no hum, no artifacts), `/tmp/audiotest_last.wav` successfully dumped
- [x] Filter library unit tests cover individual HPF/LPF/notch attenuation targets and the full POTS chain end-to-end
- [ ] Place an actual peer-to-peer call between the two phones (not possible solo)

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -1048,8 +1048,11 @@ func main() {
 		}
 		slog.Info("audio test: captured, playing back", "samples", len(recorded), "duration_s", float64(len(recorded))/float64(sampleRate), "peak", maxAmp)
 
-		if err := writePCMWav("/tmp/audiotest_last.wav", recorded, sampleRate); err != nil {
-			slog.Warn("audio test: wav dump failed", "error", err)
+		const dumpPath = "/tmp/audiotest_last.wav"
+		if err := writePCMWav(dumpPath, recorded, sampleRate); err != nil {
+			slog.Warn("audio test: wav dump failed", "path", dumpPath, "error", err)
+		} else {
+			slog.Info("audio test: wav dumped", "path", dumpPath)
 		}
 		mixer.PlayOnceSamples(recorded)
 		time.Sleep(100 * time.Millisecond)
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/binary"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


justinlindh:
@claude

justinlindh:
Thanks @claude, addressed in `ebcdc47`:

**`BiquadChain.Process` aliasing** — fixed, always returns an independent copy now. The nil/empty path does a `make+copy` to match the non-nil path. Added tests asserting non-aliasing for both the nil and empty-chain cases (mutates `out[0]` and verifies `in[0]` is untouched). You're right that the asymmetry was a foot-gun — the RNNoise call site immediately downstream mutates in place, and the invariant would have broken the first time someone added another upstream stage.

**`f.Close` swallowing errors** — already fixed in `c5267b4` before your review came in (errcheck caught the same thing). `writePCMWav` now uses the named-return-plus-deferred-close pattern and surfaces the close error when the write path hasn't already failed.

**Log wav dump path at Info** — done. `*#TEST#` now logs `"audio test: wav dumped" path=/tmp/audiotest_last.wav` after a successful dump so it shows up in `journalctl -u digitsd` alongside the existing capture log line.

**Self-test no longer exposes raw mic** — not addressing in this PR, but your suggestion is noted and I agree. The path forward when we need it again is a second service code (`*#7283#` / `*#RAW#`) that forces `Denoise=false` + dumps to a separate wav. Cheap to add when the next hardware-noise investigation comes up. Filed mentally under "follow-up when needed" rather than YAGNI-deferred — we'll definitely want it again.

**Monotonic filenames for repeated `*#TEST#`** — not taking. In practice I always `scp` between invocations; if that ever changes I'll revisit. Overwriting is self-cleaning on tmpfs which is what I want.

**60-biquad cost** — flagged as a measurement item for whoever flips `Bandpass` to true. Not blocking on it now because the default is off and the follow-up branch (`feat/digitsd-pots-character`) uses the much lighter 4-biquad `NewPOTSCharacterChain` for the cosmetic POTS voice color, applied after RNNoise instead of in front of it. Will benchmark there if we ever enable it in a live call.

**`make` cap nit** — `64` still covers `60` with a small margin and avoids a reallocation. Left as-is.

Ready for another look.

PR #166, opened 2026-04-13
feat: per-line voice_style setting with live hot-reload

justinlindh:
## Summary

Adds a per-line `voice_style` setting (`copper` | `modern`) editable in the web UI, persisted in the server DB as a JSONB column, pushed to each digitsd device over its existing signaling WebSocket, and applied live on the audio pipeline via an atomic chain swap without restarting the daemon or dropping the call.

- **Default is `copper`** (POTS-colored audio via a 300-3400 Hz 4th-order bandpass applied after RNNoise). Rationale: Digits runs inside gutted vintage desk phones, and POTS voice coloring matches the physical aesthetic.
- **Editable per line** via a new dropdown in the existing `phone-edit-row` form on `/phones`. Inline for now; refactor to a dedicated settings panel when a second per-line setting lands.
- **Authoritative server, caching device.** Settings live in `lines.settings` JSONB; digitsd caches them in `/data/digits/config.json` so cold boots without the server pick up the last-known value, and the server pushes the fresh state on every reconnect so drift is impossible.
- **Hot-reload mechanism**: `Pipeline.character` is now an `atomic.Pointer[BiquadChain]`, and `SetVoiceStyle(string)` atomically swaps it. `captureLoop` does a per-frame `Load()`, so mid-call toggles take effect on the next audio frame without dropping or glitching the call. The old chain is GC'd once the current frame finishes processing.

## Architecture

```
web UI (dropdown)
   │
   ▼
POST /phones/<num>/edit
   │
   ├─ line.Store.UpdateSettings  (DB: JSONB column on lines)
   │
   └─ Hub.SendTo(TypeLineSettings)  (WebSocket, best-effort)
           │
           ▼
   digitsd signal inbox
           │
           ├─ Pipeline.SetVoiceStyle  (atomic.Pointer swap on live pipeline)
           │
           └─ Config.Save  (atomic tmp+rename to /data/digits/config.json)
```

Server also pushes `TypeLineSettings` immediately after every successful `Hub.Register`, so reconnects always pull fresh authoritative state.

## Notable design choices

- **JSONB column over a separate settings table.** First per-line setting of what will be many; JSONB means no schema migration per new setting, and the Go layer layers `DefaultSettings().Merge(patch).Normalize()` to turn empty DB rows into default-filled values.
- **`signaling.LineStore` interface + `lineStoreAdapter`.** Keeps the signaling package a leaf — only the adapter file imports `internal/line`, matching how `CallAuthorizer` is handled.
- **Three copies of the `"copper"` / `"modern"` constants** across `config`, `line`, and `signaling` packages (the `audio` package's copies were removed during simplify cleanup). This is intentional: each package at a different boundary needs its own vocabulary, and there's no shared leaf package worth creating for two strings.
- **`newPipeline()` helper** on `daemonCallbacks` DRYs the two call-setup sites (`InitiateCall`, `AnswerCall`) so adding a second config field that controls the pipeline only touches one place.
- **Idempotent `TypeLineSettings` handler** skips the config save when the incoming style already matches the cached value, avoiding SD-card fsync on every reconnect.

## Commits (17 total, squash-merges cleanly)

12 plan-driven feature commits (`9a73ccb` → `7bb44b8`), three post-review fix commits (`7e1fb66` race fix, `4a08351` normalization, `19583dc` simplify cleanups), and three branch-tail commits from earlier session work on the POTS character filter primitive (`54281e6` + add/revert pair) that introduce the filter this PR toggles.

## Test plan

- [x] `go build`, `go vet`, `golangci-lint run`, `go test` — all clean on both `server/` and `pi/digitsd/` modules
- [x] TDD for all new types: `line.Settings` (defaults, merge, normalize, JSON round-trip), `Pipeline.SetVoiceStyle` (toggle, unknown-fallback), `config.VoiceStyleOrDefault`, `line.Store` round-trip with empty-JSON fallback to default
- [x] Signaling wire format: `TestLineSettingsMessageJSONRoundTrip` exercises the full Message round trip
- [x] Web handler: test fixtures updated, existing tests pass
- [ ] **Manual on-device verification** (to run after merge or before depending on your appetite):
  - [ ] Deploy new digitsd to both phones, deploy new signald to prod
  - [ ] Verify migration v11 ran: `\d lines` shows `settings | jsonb`
  - [ ] Confirm both phones log \`line_settings applied voice_style=copper\` on reconnect
  - [ ] Dial \`*#8378#\` on digits 2, confirm copper sound
  - [ ] Flip digits 2 to modern via web UI, dial \`*#8378#\` again, confirm modern sound
  - [ ] Reboot digits 2, confirm setting survived (cached in config.json)
  - [ ] Flip digits 2 back to copper

## Known out-of-scope issues

- Pre-existing unsynchronized writes to `Config.DeviceToken` and `Config.PairingCode` in the `TypePaired` handler. Same unsafe pattern as the race this PR fixed for `VoiceStyle`, but not a regression; should be addressed in a small follow-up that unifies the three.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ node_modules/
 
 # Superpowers
 .superpowers/
+docs/superpowers/
 
 # Claude Code
 .claude/
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -199,9 +199,7 @@ func (d *daemonCallbacks) InitiateCall(targetNumber string) {
 	close(sdpSent)
 
 	// Start audio pipeline
-	pipCfg := audio.DefaultPipelineConfig()
-	pipCfg.Character = d.cfg.VoiceStyleOrDefault() == config.VoiceStyleCopper
-	d.pipeline = audio.NewPipeline(pipCfg)
+	d.pipeline = d.newPipeline()
 	if err := d.pipeline.Start(); err != nil {
 		slog.Error("audio pipeline start failed", "error", err)
 		return
@@ -376,9 +374,7 @@ func (d *daemonCallbacks) AnswerCall() {
 	close(sdpSent)
 
 	// Start audio pipeline (capture only — playback goes through mixer)
-	pipCfg := audio.DefaultPipelineConfig()
-	pipCfg.Character = d.cfg.VoiceStyleOrDefault() == config.VoiceStyleCopper
-	d.pipeline = audio.NewPipeline(pipCfg)
+	d.pipeline = d.newPipeline()
 	if err := d.pipeline.Start(); err != nil {
 		slog.Error("audio pipeline (answer) start failed", "error", err)
 		return
@@ -438,11 +434,19 @@ func (d *daemonCallbacks) HangupCall() {
 	slog.Info("call ended")
 }
 
-// applyVoiceStyleLive forwards a voice style change to the active audio
-// pipeline if one is running. Called from the signal inbox handler when a
-// line_settings update arrives; safe to call with no active pipeline (it
-// becomes a no-op, and the config cache still gets persisted so the next
-// call picks up the style at construction time).
+// newPipeline constructs a fresh capture pipeline with per-line config
+// applied (voice style, etc). The returned pipeline is not yet started.
+// Must be called with d.mu held (reads d.cfg without acquiring the lock).
+func (d *daemonCallbacks) newPipeline() *audio.Pipeline {
+	cfg := audio.DefaultPipelineConfig()
+	cfg.Character = d.cfg.VoiceStyleOrDefault() == config.VoiceStyleCopper
+	return audio.NewPipeline(cfg)
+}
+
+// applyVoiceStyleLive forwards a voice style change to the active pipeline
+// if one is running. Safe to call with no active pipeline: it becomes a
+// no-op, and the config cache save in setVoiceStyleConfig still persists
+// the style for the next call to pick up at construction time.
 func (d *daemonCallbacks) applyVoiceStyleLive(style string) {
 	d.mu.Lock()
 	pip := d.pipeline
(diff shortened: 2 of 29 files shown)


justinlindh:
@claude

justinlindh:
Thanks @claude. Addressed three, skipping three with reasons. Pushed as `9d0383f`.

**Fixed:**

1. **Misleading `linestore.go` comment** — reworded to describe the actual pattern ("dependency on internal/line is isolated to a single adapter file") rather than the false claim that signaling is dependency-free.

3. **`rawDB` anonymous interface** — replaced with plain `*sql.DB`. Simpler, and `database/sql` is already imported.

4. **Canonical-constant cross-refs on both `protocol.go` `LineSettings` structs** — added a short note pointing at `line.VoiceStyle*` and `config.VoiceStyle*` as the authoritative list, with a checklist for contributors adding new styles. Cheap insurance against drift.

**Skipping:**

2. **`pushLineSettings` pre-flight `hub.Get(number) == nil`** — keeping it. `Hub.SendTo` returns `fmt.Errorf("phone %s not connected", number)` for the offline case, and the caller in `handlePhoneEditPost` logs that at `slog.Warn`. Without the pre-flight, every settings edit made while the line's device happens to be offline (the common case if you're editing a kid's phone while they're asleep) fires a warn log. The pre-flight converts that into a silent no-op which is the right default for the offline-is-normal case. Not belt-and-suspenders — it's changing "warn-loud on expected offline" to "quiet on expected offline, still warn on real send failures".

5. **Loosen passband tolerance to ±1.5 dB** — not changing. The filter is deterministic and the math at the test frequencies is well inside ±1 dB: 1800 Hz through the cascaded 4th-order 3400 Hz LPF is roughly -0.66 dB, and 800 Hz through the 300 Hz HPF skirt is under -0.2 dB. No platform-specific flake risk because biquad state is pure float64 arithmetic on aligned inputs. If it ever does flake I'll revisit, but the current bound is expressing the design intent, not accidentally tight.

6. **Form `voice_style == ""` silent skip** — acknowledged as a future concern, but no action today because the field is a hard-coded two-value select in the template and a missing field means "template regression, don't touch anything" rather than "user wants to clear". When a third field lands and the "no change" semantics get interesting, we'll revisit.

CI is running on `9d0383f`. On-device manual verification from the test plan still pending.

PR #168, opened 2026-04-13
feat(server): move voice_style control to phone detail page

justinlindh:
## Summary
- Removes the Voice dropdown from the /phones list inline edit row
- Adds a Voice Style section to the /phones/{number} detail page's Details card, styled to match the existing settings page selects (native control, matching focus ring and button)
- New POST /phones/{number}/voice-style handler updates line settings, pushes live to the connected device, and htmx-swaps just the voice-style section on success

## Test plan
- [x] Verified in prod via agent-browser: switch Copper wire → Modern HD → save → DB reflects \`{"voice_style": "modern"}\`, section re-renders with Modern HD selected
- [x] Revert to Copper wire round-trip works the same
- [x] go vet + go test ./internal/web/... pass

### DIFF
--- a/server/internal/web/handler.go
+++ b/server/internal/web/handler.go
@@ -723,13 +723,18 @@ func (h *Handler) handlePhoneVoiceStylePost(w http.ResponseWriter, r *http.Reque
 		http.Error(w, "bad request", http.StatusBadRequest)
 		return
 	}
+	raw := strings.TrimSpace(r.FormValue("voice_style"))
+	if raw == "" {
+		http.Error(w, "missing voice_style", http.StatusBadRequest)
+		return
+	}
 	ln, err := h.lineStore.GetByNumber(number)
 	if err != nil {
 		http.NotFound(w, r)
 		return
 	}
 	next := ln.Settings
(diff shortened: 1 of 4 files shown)


justinlindh:
Final UI on prod:

voice style options card

justinlindh:
@claude

justinlindh:
Addressed #1 in 545137b — `handlePhoneVoiceStylePost` now returns 400 on an empty `voice_style` instead of letting `Normalize()` silently coerce to copper.

Skipping #2 (matches surrounding convention), #3 (explicit not-worth-it), and #4 (cosmetic, low priority) per your own notes.

justinlindh:
Reconsidered #2 and added tests in 2fd4ccd. Seven cases covering the real behavior contracts:

- empty form → 400 (guards the Normalize silent-downgrade)
- valid change → DB write + 303 redirect
- HTMX → partial render with correct radio checked
- unknown value → Normalize coerces to copper
- missing line → 404
- live push: connected device receives a `TypeLineSettings` message with the new value
- no-op (same value) save → no push at all (idempotency guarantee from the earlier simplify review)

Skip-on-no-`TEST_DATABASE_URL` to match the rest of handler_test.go. All seven pass against a local Postgres.

PR #171, opened 2026-04-13
test(server): split unit and integration tests with //go:build tag

justinlindh:
## Summary
- Introduces a `//go:build integration` tag on every server test file that requires Postgres so the unit suite stops silently skipping DB-backed tests. 21 test files across `internal/{admin,auth,calls,db,device,household,line,pairing,web}` are tagged.
- Adds a `## Testing` section to `CLAUDE.md` documenting the unit/integration/e2e taxonomy, build tag convention, and local commands.
- Renames the existing `server-ci.yml` job from `test` to `unit` for clarity in the Actions UI (no behavioral change).
- Adds a new `.github/workflows/server-integration.yml` that provisions two postgres service containers (signald + admind) and runs `go test -tags=integration -race -p 1 ./...`.
- Fixes every integration test that was silently broken before the split. See "Test fixes" below.

## Why
CI was running `go test ./...` without `TEST_DATABASE_URL` set, so every DB-backed test silently skipped. The green check on PRs told you the code compiled and the handful of pure unit tests passed; it told you nothing about SQL code, handler integration, or migration behavior. After this PR the unit job is a real fast gate (~1.5s) and the integration job actually exercises the database layer.

Exposing the previously-skipped tests revealed 19 failures of varying vintage. Rather than merging broken tests into a dedicated `integration` job, this PR fixes them all so the new workflow goes green on first run.

## Test fixes

**Deleted** (pure Postgres test with zero app signal):
- `internal/db/db_test.go::TestOpenAndMigrate` — introspected `information_schema.tables` against a hardcoded list of table names. If migrations fail, `Open()` already returns an error, and if a table is missing, every other integration test that touches it fails first. The hardcoded list had also bit-rotted (v6 renamed `phones` to `lines` and dropped `phones`). Deleted the test and its wrapper file rather than patching the stale list.

**Fixed in place** (all seeded against real rows):
- `internal/pairing/pairing_test.go` — `TestClaimDevice_Success`, `TestClaimDevice_FailOnInvalidCode`, `TestClaimDevice_FailOnExpiredCode` passed a hardcoded dummy household UUID that never existed; the `lines.household_id` FK rejected it. Added a `seedHousehold` helper that inserts a real household row. `TestCleanupExpiredCodes` inserted a device with `line_id = 0` (non-existent, FK violation); migration v7 already made the column nullable because a device exists before it's paired, so the correct fixture is `line_id = NULL`.
- `internal/web/handler_test.go` — 13 tests all redirected 303 to `/onboard` because `addSessionCookie` created a user but never inserted `household_members`. New `setupAuthedHousehold` helper combines session creation with household seeding. The four `TestPhoneRestart*` tests also inserted lines with `household_id = 'h1'` (invalid UUID, FK silently dropped via `ON CONFLICT DO NOTHING`); they now use the real household id. `TestDeletePhone` had the same issue with a zero UUID in `lineStore.Add`. `TestCallsPageReturns200` additionally enables call history on the seeded household because `handleCalls` redirects to `/settings` when it is disabled. `TestDashboardReturns200` and `TestSettingsPageReturns200` had stale content assertions (`"Network Overview"` only shows without a household, `"SIGNALD_ADDR"` is not in the settings template) replaced with content that is actually rendered.
- `internal/web/e2e_test.go` — `TestE2EFullCallFlow` and `TestE2EWebUIWithData` inserted lines with the same dummy zero UUID; replaced with a `seedE2EHousehold` helper. `setupTestServer` now wires a real `householdStore` into `NewHandler` so the `/api/status` call in `TestE2EFullCallFlow` clears the onboarding middleware. Separately, the WS register handler was tightened in an earlier PR to require `hardware_id`; the register calls now send stable hardware ids.
- `internal/web/e2e_calls_test.go` — `TestCallsPageScopedToHousehold` had two latent bugs: `handleCalls` redirects to `/settings` when call history is disabled (the test was asserting content from `/settings`, not `/calls`), and the calls template formats phone numbers through `line.FormatNumber` (`5550001` renders as `555-0001`). Enabled call history on household A and asserted against the formatted numbers.

## CI workflow correctness fix
`server-integration.yml` uses `go test -tags=integration -race -count=1 -p 1 ./...`. The `-p 1` serializes package execution because each package's `Open()` runs migrations on the shared Postgres, and parallel `CREATE TABLE IF NOT EXISTS` statements can race. Wall-clock impact is minimal (individual package runtime is sub-second).

## Verification performed locally
- `go test ./...` from `server/` — every package `ok` or `[no test files]`, 1.4 seconds, zero `SKIP` lines referencing `TEST_*_DATABASE_URL`.
- `go test -tags=integration -race -count=1 -p 1 ./...` against a postgres:16-alpine container with `digits_test` and `digits_admin_test` databases — every package `ok`, 25 seconds.
- `go vet ./...` and `go vet -tags=integration ./...` clean.

## Branch protection

Done via gh api (ruleset 14842091). `unit` and `integration` are now required status checks on `main`. Two other open PRs (#169, #170) do not touch `server/` and will not produce these checks under the current `paths:` filter on the workflows; they will need an admin bypass or a trivial rebase touching `server/**` or the workflow files to satisfy the gate once this PR merges.


justinlindh:
@claude

PR #174, opened 2026-04-13
ci(pi): add PR test workflow for digitsd and digits-setup

### LINKED ISSUE #173: Add PR-level unit test CI for pi/ modules
## Background

The `.github/workflows/golangci-lint.yml` workflow runs golangci-lint on PRs that touch `pi/digitsd/**` or `pi/digits-setup/**`, but only lint, not `go test`. The `pi-release.yml` workflow only runs at tag time (push to main with `pi/**` paths) and delegates to semantic-release.

Result: a PR that breaks a digitsd or digits-setup unit test can merge cleanly as long as it lints. The break is only caught when main is built for release, after the bad code has already landed.

## Why it matters

This came up during PR #170 while adding the wifi auto-fallback supervisor and 40+ new tests. We were running the full suite locally on every commit, so the branch is clean. But for future contributors (including me) it's a real gap: tests exist but nothing in CI enforces that they stay green on PRs.

## Proposed scope

Add a new workflow `pi-ci.yml` (or add test jobs to the existing lint workflow) that:

1. Triggers on PRs touching `pi/digitsd/**` or `pi/digits-setup/**`.
2. Runs `go test ./...` in `pi/digitsd/` with the C deps it needs (libasound2-dev, libopus-dev, libopusfile-dev — same setup as `lint-digitsd`).
3. Runs `go test ./...` in `pi/digits-setup/` (pure Go, no deps).
4. Runs `go test -race ./internal/wififallback/...` specifically (race-detector coverage for the supervisor).
5. Uses the same Go version as `pi-release.yml` (currently 1.26.1) and caches go.sum dependencies.

Optional follow-on: also add `go build ./cmd/digitsd` as a compile check (would catch import-level breakage that tests alone miss).

## Out of scope

- Cross-compiling to arm64 in CI. The existing `pi-release.yml` handles that at release time; duplicating in PR CI would roughly double the runtime for limited benefit.
- Integration tests on real hardware. Those are manual by nature and tracked in the PR description.

## Related

- PR #170 (wifi auto-fallback) flagged this as an out-of-scope observation.

justinlindh:
Closes #173.

## Summary
- New `.github/workflows/pi-ci.yml` runs `go build` + `go test ./...` for `pi/digitsd` and `pi/digits-setup` on PRs touching those paths.
- `test-digitsd` installs the same C deps as `lint-digitsd` (libasound2-dev, libopus-dev, libopusfile-dev) and ends with `go test -race ./internal/wififallback/...` for supervisor race coverage.
- Go 1.26.1 pinned to match `pi-release.yml` and `golangci-lint.yml`; per-module `go.sum` caching.
- Cross-compile and hardware integration intentionally excluded per the issue's out-of-scope section.

## Test plan
- [x] `go build ./... && go test ./...` in `pi/digitsd`
- [x] `go test -race ./internal/wififallback/...`
- [x] `go build ./... && go test ./...` in `pi/digits-setup`
- [ ] CI green on this PR
- [ ] Add `test-digitsd` and `test-digits-setup` to required status checks for `main` after merge

### DIFF
--- a/.github/workflows/pi-ci.yml
+++ b/.github/workflows/pi-ci.yml
@@ -1,0 +1,55 @@
+name: Pi CI
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'pi/digitsd/**'
+      - 'pi/digits-setup/**'
+      - '.github/workflows/pi-ci.yml'
+  pull_request:
+    paths:
+      - 'pi/digitsd/**'
+      - 'pi/digits-setup/**'
+      - '.github/workflows/pi-ci.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  test-digitsd:
+    name: test-digitsd
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version: '1.26.1'
+          cache-dependency-path: pi/digitsd/go.sum
+      - name: Install C dependencies
+        run: sudo apt-get update && sudo apt-get install -y libasound2-dev libopus-dev libopusfile-dev
+      - name: Build digitsd
+        working-directory: pi/digitsd
+        run: go build ./...
+      - name: Test digitsd
+        working-directory: pi/digitsd
+        run: go test ./...
+      - name: Race test wififallback supervisor
+        working-directory: pi/digitsd
+        run: go test -race ./internal/wififallback/...
+
+  test-digits-setup:
+    name: test-digits-setup
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version: '1.26.1'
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the review.

1. **`pi/digits-setup/go.sum` missing** — confirmed the file doesn't exist, but the first CI run on this PR went green (`Pi CI` / `test-digits-setup` passed in run 24337680450), so `setup-go@v6` handles the missing glob without erroring in this case. Leaving as-is; if it ever flakes I'll drop `cache-dependency-path` for that job.
2. **Double-run of wififallback under non-race + race** — intentional and matches the issue scope (race coverage targeted to the supervisor only). Keeping the targeted form.
3. **`timeout-minutes` / `concurrency`** — agreed these are nice-to-haves but absent from the sibling pi workflows. Skipping for consistency; happy to add in a follow-up that does it across all pi/server workflows together.

No code changes from this review.

PR #175, opened 2026-04-13
feat(pi): write nmconnection files directly to /etc/NM, drop reboot-on-configure

### LINKED ISSUE #172: Write nmconnection files directly to /etc/NetworkManager/system-connections/
## Background

PR #170 (wifi auto-fallback) introduced per-SSID `.nmconnection` files written by `digits-setup`'s captive portal, with `digits-ap-check` copying them from `/data/wifi/` to `/etc/NetworkManager/system-connections/` on boot. The copy path is what made the PR #170 fixup necessary (glob-aware script, stale-file pruning, etc).

During review, @claude suggested eliminating the copy step entirely by having `configure.go` write directly to `/etc/NetworkManager/system-connections/`. That was deferred as a follow-up.

## Why it's worth doing

- Eliminates the `/data/wifi/` → `/etc/NM/...` copy coupling. Today, a user configuring WiFi writes to `/data/wifi/`, then has to reboot for `digits-ap-check` to stage the file where NetworkManager actually looks. That reboot is the reason configure currently ends with a `schedule-reboot` call.
- Removes `digits-ap-check`'s per-SSID copy logic and the stale-file pruning we just added.
- Would let a user reconfigure WiFi without rebooting: NM notices a new connection file via inotify and picks it up immediately.
- Decouples provisioning from the boot script entirely, which simplifies recovery and factory-reset semantics.

## Why it wasn't done in PR #170

- The captive portal process runs as the `digits` user with no rootfs-rw privileges by default. Writing to `/etc/` requires either remounting `/` read-write or giving `digits-setup` a sudoers entry for specific paths.
- `/data/` remains writable even when rootfs is ro, which is why the current design uses it as the source of truth.
- A move to direct writes would also want to move the \"source of truth\" concept: do we still keep a copy in `/data/wifi/` for backup/factory-reset, or is `/etc/NM/system-connections/` canonical?

## Proposed scope

1. Add a sudoers entry for `digits-setup` allowing remount rw/ro of `/` and cp/chmod into `/etc/NetworkManager/system-connections/`.
2. Update `pi/digits-setup/internal/wifi/configure.go` to remount, write directly, remount ro, and drop the reboot scheduling on success (NM should pick up the new connection automatically).
3. Remove the per-SSID copy and prune logic from `digits-ap-check`'s `do_station_up`. Leave `do_init`'s decision about station-vs-AP intact (still needs the `/data/wifi-configured` flag).
4. Decide whether to keep a backup copy in `/data/wifi/` or remove that directory entirely.
5. Re-test multi-network roam and factory-reset flows.

## Related

- PR #170 (wifi auto-fallback + multi-network memory)

justinlindh:
## What

Closes #172. `digits-setup` now dual-writes each `.nmconnection` atomically to
`/etc/NetworkManager/system-connections/` (operational store NM reads live)
and `/data/wifi/` (persistent backup that survives rootfs replacement on OTA
image upgrades), then defers an `digits-ap-check down` call by 2s so the
captive portal response flushes to the client before hostapd stops.
`digits-ap-check` stops hot-copying on every boot; `do_init` only restores
from `/data/wifi/` when the operational store is empty (the OTA cold path).

## Why

PR #170 introduced per-SSID `.nmconnection` files but required a reboot after
configure so `digits-ap-check` could stage them into `/etc/NM/...`. Writing
directly removes the reboot, lets users reconfigure Wi-Fi without rebooting,
and deletes the per-boot copy/prune logic. `/data/wifi/` stays as a backup so
OTA image rebuilds do not factory-reset every device's Wi-Fi config.

### Automated
- [x] `cd pi/digits-setup && go build ./... && go vet ./... && go test ./... && golangci-lint run ./...`
- [x] `bash -n pi/image/rootfs-overlay/usr/local/bin/digits-ap-check`
- [x] `cd pi/digitsd && go build ./...`

### Manual (on both phones, post-merge)
- [ ] Fresh AP-mode device: configure via captive portal, verify HTTP 200 reaches the browser before AP drops, device joins station mode without rebooting
- [ ] Already-online device: reconfigure to a different network, verify NM picks it up (if not, follow-up to add `nmcli reload && nmcli up`)
- [ ] Simulated OTA: `rm /etc/NetworkManager/system-connections/digits-wifi*.nmconnection` on a working device, reboot, verify cold-path restore from `/data/wifi/`
- [ ] Inject a null-byte file into `/etc/NM/.../digits-wifi-bad.nmconnection`, reboot, verify fallback
- [ ] Fallback supervisor triggers `reset_wifi_config`, verify both dirs are wiped

### DIFF
--- a/pi/digits-setup/cmd/digits-setup/main.go
+++ b/pi/digits-setup/cmd/digits-setup/main.go
@@ -17,8 +17,9 @@ func main() {
 
 	scanner := &wifi.SystemScanner{}
 	configurator := &wifi.SystemConfigurator{}
+	ap := wifi.SystemAPController{}
 
-	mux := portal.NewHandler(scanner, configurator)
+	mux := portal.NewHandler(scanner, configurator, ap)
 
 	log.Printf("digits-setup listening on %s", addr)
 	if err := http.ListenAndServe(addr, mux); err != nil {
--- a/pi/digits-setup/internal/portal/handler.go
+++ b/pi/digits-setup/internal/portal/handler.go
@@ -3,9 +3,11 @@ package portal
 import (
 	"embed"
 	"encoding/json"
+	"errors"
 	"io/fs"
 	"log"
 	"net/http"
+	"sync/atomic"
 	"time"
 
 	"github.com/justinlindh/digits/pi/digits-setup/internal/wifi"
@@ -14,9 +16,20 @@ import (
 //go:embed static/*
 var staticFiles embed.FS
 
+// apTeardownDelay defers `digits-ap-check down` after a successful configure
+// so two things can happen in order: (1) the HTTP response bytes leave wlan0
+// before hostapd stops, and (2) the client has a moment to read them before
+// losing AP association.
+const apTeardownDelay = 2 * time.Second
+
 // NewHandler returns the HTTP mux for the captive portal.
 func NewHandler(scanner wifi.Scanner, configurator wifi.Configurator, ap wifi.APController) http.Handler {
 	mux := http.NewServeMux()
+
+	// teardownScheduled gates the deferred AP teardown so a duplicate or
+	// retried /api/configure cannot stack multiple `digits-ap-check down`
+	// goroutines.
+	var teardownScheduled atomic.Bool
 
 	// Captive portal detection — redirect to setup page
 	captiveRedirect := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -60,7 +73,11 @@ func NewHandler(scanner wifi.Scanner, configurator wifi.Configurator, ap wifi.AP
 
 		if err := configurator.Configure(req); err != nil {
 			log.Printf("configure error: %v", err)
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			status := http.StatusInternalServerError
+			if errors.Is(err, wifi.ErrInvalidRequest) {
+				status = http.StatusBadRequest
+			}
+			http.Error(w, err.Error(), status)
 			return
 		}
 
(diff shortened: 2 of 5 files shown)


justinlindh:
@claude

justinlindh:
Addressed the review in 0cd3f50.

**Concerns**

1. **`log()` stdout contamination** — fixed. `log()` now writes to stderr + `/dev/kmsg` only; `list_valid`'s captured stdout stays clean so a corrupt backup file alongside a valid one no longer contaminates the cold-path restore loop.
2. **`/api/configure` status codes** — added `wifi.ErrInvalidRequest` sentinel. The handler now returns 400 only when `errors.Is(err, wifi.ErrInvalidRequest)` is true, 500 for mount/filesystem failures.

**Smaller notes**

- Named const `apTeardownDelay = 2 * time.Second` with a comment covering both reasons for the delay.
- Deferred AP teardown gated by `atomic.Bool` (`CompareAndSwap`) so duplicate `/api/configure` calls don't stack `ap.Down()` goroutines.
- Flag file permissions tightened from 0644 to 0600.
- Added a comment at the legacy cleanup loop noting the single hardcoded filename is deliberate.
- Added a comment on `uuidForSSID` clarifying it's a stable identifier, not RFC 4122.
- New test `TestConfigureOperationalWriteFailureKeepsBackupOmitsFlag` locks in the "half-written config never promotes to station" invariant: op-write fails → backup remains, flag absent, `roCalls=1` (defer ran). Also added `TestConfigureMissingSSIDIsInvalidRequest` to pin the sentinel.

`go build ./... && go vet ./... && go test ./... && golangci-lint run ./...` clean. `bash -n digits-ap-check` clean.

justinlindh:
@claude - this good now?

PR #177, opened 2026-04-13
fix(digitsd): self-heal corrupted or unwritable device-id

justinlindh:
## Summary

Digits 1 went offline after a reboot today and got stuck in a tight reconnect loop, rejected by the server on every WS register with `hardware_id required`. Root cause: `/data/digits/device-id` was 37 NUL bytes. Two compounding bugs in digitsd let that state both exist and persist across restarts.

**Bug 1 — brittle validation.** `LoadOrCreateDeviceID` accepted any 36-byte content after `strings.TrimSpace`. TrimSpace does not strip NULs, and a raw length check does not verify UUID shape. Fix: `strings.Trim` over `" \t\r\n\x00"` and a proper UUID v4 shape check.

**Bug 2 — self-heal impossible.** The regen path used `os.WriteFile`, which cannot open an existing file owned by another user. `digits-first-boot.sh` runs as root and created `/data/digits/device-id` as root:root, but digitsd runs as the `digits` user, so the regen branch failed with EACCES on every legacy image. Fix: `os.Remove` before `os.WriteFile` (digits user has write perms on the parent directory, so the unlink succeeds). Also chown the file to `digits:digits` at first-boot so new images never need the self-heal path.

Digits 1 was already restored out of band by manually writing the paired UUID from the `devices` table back to `/data/digits/device-id`.

## Test plan

- [x] `go test ./internal/config/...` — new cases cover valid file, missing file, 36-byte NUL-filled file, and unwritable file owned by another user (simulated via 0444)
- [x] `go test ./...` across digitsd — all green
- [x] `go vet ./...`, `golangci-lint run ./...` — clean
- [ ] Deploy to digits 1 and digits 2 and verify both stay online through a reboot

### DIFF
--- a/pi/digitsd/internal/config/deviceid.go
+++ b/pi/digitsd/internal/config/deviceid.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
@@ -13,10 +14,14 @@ const deviceIDPath = "/data/digits/device-id"
 // If the file doesn't exist or contains a non-UUID value (legacy 4-char hex),
 // generates a new UUID v4 and persists it.
 func LoadOrCreateDeviceID() (string, error) {
-	data, err := os.ReadFile(deviceIDPath)
+	return loadOrCreateDeviceIDAt(deviceIDPath)
+}
+
+func loadOrCreateDeviceIDAt(path string) (string, error) {
+	data, err := os.ReadFile(path)
 	if err == nil {
-		id := strings.TrimSpace(string(data))
-		if len(id) == 36 {
+		id := strings.Trim(string(data), " \t\r\n\x00")
+		if isUUIDv4Shape(id) {
 			return id, nil
 		}
 	}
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude please review

justinlindh:
Addressed in ba61c4a. Point by point:

1. **Variant nibble check** — added. Position 19 now requires `8/9/a/b` (or uppercase). Consistency with the `'4'` check at position 14.
2. **`range` over string** — switched `isUUIDv4Shape` to `for i := 0; i < 36; i++ { c := s[i] }`. You're right it was correct by accident; explicit byte indexing is clearer.
3. **Non-atomic replace** — switched to `os.CreateTemp` + `os.Rename`. Rename handles the cross-user case equally well (only needs dir write perms) and eliminates the power-loss window. Added a test assertion that the regen path leaves no stray temp files behind.
4. **Stale godoc** — updated. Now describes the real behavior (missing / unreadable / non-v4 / NUL-filled all trigger regen) instead of the legacy-4-char-hex story.
5. **Test uses `isUUIDv4Shape`** — done in both `CorruptedFixedLengthFile` and `UnwritableExistingFile`. Also added a focused `TestIsUUIDv4Shape` table test covering valid, uppercase, wrong length, wrong version nibble, wrong variant nibble, non-hex, missing hyphen, all-NULs, and empty.
6. **`/data/digits` parent ownership** — checked, no change needed. `pi/image/init-data.sh:70` creates the directory at image build time with `install -d -m 755 -o 999 -g 992`, which writes `digits:digits` directly to the filesystem image. All existing fielded devices already show `drwxr-xr-x digits digits` on `/data/digits`, verified on digits 1 just now. The `mkdir -p /data/digits` in `digits-first-boot.sh` is a defensive no-op that happens after the dir already exists.

PR #183, opened 2026-04-19
fix(image): clear boot counter in AP mode to prevent false recovery

justinlindh:
## Summary

- The initramfs boot counter was only cleared by `digitsd`, which is stopped during AP mode. On a fresh flash, power-cycling 3 times before configuring WiFi would trip the recovery threshold.
- Clear the boot counter in `digits-ap-check` for both `do_ap_up()` and `do_station_up()` -- if the device reached that service, the boot was successful.

## Test plan

- [ ] Flash a fresh image, verify AP mode starts
- [ ] Power-cycle 3+ times without configuring WiFi, verify AP mode still starts (not recovery)
- [ ] Configure WiFi, reboot, verify station mode works normally


justinlindh:
@claude

PR #192, opened 2026-04-20
fix(digitsd): play POTS off-hook treatment after dial-tone timeout

justinlindh:
## What

When the user picks up the handset and lets dial tone play past the firmware's dial-tone timeout, the Pico FSM transitions to BUSY and silently drops keypresses. The Pi never learns about the timeout (\`TIMEOUT:DIAL_TONE\` was unhandled), keeps streaming dial tone audio, and the keypad appears dead until the user cycles the hook.

This PR wires the Pi controller to handle \`TIMEOUT:DIAL_TONE\` by running the same 90s POTS off-hook permanent-signal treatment (Bellcore GR-506-CORE) the codebase already implements for remote hangup: 1s silence → 45s reorder → 3min howler → lockout silence. Also restores the bench-test 30s timeout to a production-accurate 15s, and ships a pile of code-review fallout from auditing the change.

## Why

The keypad-dead-after-timeout symptom is the visible bug. Underneath, the firmware's BUSY state was meant to signal "off-hook, no longer accepting input" but the Pi never received that signal, so the user got no audible cue (dial tone kept playing) and no functional cue (keys stopped working). 90s POTS networks handled this with the permanent-signal treatment escalation, which is the behavior we want this private phone network to mimic.

## What's in the diff

**Bug fix proper:**
- \`controller.go\`: new \`StateOFFHOOK_TIMEOUT\`, new \`onTimeoutDialTone\` handler, shared \`runPermanentSignalTreatment\` helper extracted from \`onSignalHangup\`.
- Generation counter (\`treatmentGen\`) so a stale goroutine from a previous session can't step a fresh session forward across rapid hook-flap.
- \`Controller.Close()\` cancels in-flight sleeps so SIGTERM doesn't leave a 4-minute goroutine running.
- \`firmware/src/phone_fsm.c\`: \`DIAL_TIMEOUT_MS\` and \`DIAL_TONE_TIMEOUT_MS\` 30s → 15s.

**Surrounding cleanup (caught by review):**
- Tone names hoisted to \`phone.Tone*\` constants; \`main.go\` dispatch and the legacy socket \`TONE:\` handler now use them.
- \`mockCallbacks\` gained a mutex + snapshot accessors so tests are race-free against the controller's goroutines.
- Service-code and easter-egg tests had a dead \`_ = called\` line hiding both a race and a missing assertion; replaced with channel-based sync that actually verifies the callback fired.
- \`mixer.Stop()\` now waits for the render loop to exit (\`doneCh\` closed in renderLoop's defer) instead of returning while the goroutine may still be mid-\`WriteFrame\`. The old behavior was a real production correctness bug, not just a test issue.
- Fixed two flaky mixer tests that asserted on \`periods[0]\` after \`Start → PlayLoop/PlayOnce\` whose ordering wasn't deterministic; queue the tone before \`Start\` instead.
- \`pi-ci\` now runs the full digitsd module under \`-race\`, replacing the two narrow per-package invocations.

## Test plan

- [x] \`go test ./...\` passes (all 13 packages)
- [x] \`go test -race ./...\` passes (full module, was previously failing on \`internal/audio\` and \`internal/phone\`)
- [x] \`go vet ./...\` clean
- [x] Firmware builds (\`make firmware\`)
- [x] New \`TestController_DialToneTimeout\` covers state transition, REORDER tone, key-ignored, hook-on recovery
- [x] New \`TestController_DialToneTimeout_IgnoredOutsideDialTone\` covers defensive-no-op cases
- [x] New \`TestController_CloseAbortsOffHookTreatment\` verifies Close stops the sequence before REORDER fires
- [ ] Manual: pick up phone, wait 15s, confirm reorder tone starts; press digits, confirm they're ignored; hang up, confirm IDLE returns

### DIFF
--- a/.github/workflows/pi-ci.yml
+++ b/.github/workflows/pi-ci.yml
@@ -36,12 +36,9 @@ jobs:
       - name: Build digitsd
         working-directory: pi/digitsd
         run: go build ./...
-      - name: Test digitsd
+      - name: Test digitsd (with race detector)
         working-directory: pi/digitsd
-        run: go test ./...
-      - name: Race test wififallback supervisor
-        working-directory: pi/digitsd
-        run: go test -race ./internal/wififallback/...
+        run: go test -race ./...
 
   test-digits-setup:
     name: test-digits-setup
--- a/firmware/src/phone_fsm.c
+++ b/firmware/src/phone_fsm.c
@@ -15,8 +15,8 @@
 #include "pico/time.h"
 
 #define DIAL_DIGITS_REQUIRED 7
-#define DIAL_TIMEOUT_MS 30000  // 30s for bench testing (restore to 5000 for real phone)
-#define DIAL_TONE_TIMEOUT_MS 30000  // 30s of dial tone before off-hook timeout → busy
+#define DIAL_TIMEOUT_MS 15000        // 15s between digits before partial dial → off-hook timeout
+#define DIAL_TONE_TIMEOUT_MS 15000   // 15s of dial tone with no keys → off-hook timeout (Bellcore GR-506-CORE)
 
 static phone_state_t s_state = PHONE_STATE_IDLE;
 static char s_digits[DIAL_DIGITS_REQUIRED + 1];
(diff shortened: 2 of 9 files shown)


justinlindh:
@claude

PR #208, opened 2026-04-22
fix(server): distinguish user-not-found from real DB errors

### LINKED ISSUE #207: Distinguish 'user not found' from real DB errors in auth.Store lookups
## Context

Follow-up from PR #206 review ([thread](https://github.com/justinlindh/digits/pull/206#discussion_r3120917504)).

`auth.Store.GetUserByEmail`, `GetUserByGoogleID`, and `GetUserByID` all return `fmt.Errorf("user not found")` for `sql.ErrNoRows` and wrap every other error as the raw driver error. Callers that want to "find or create" uniformly treat *any* error as not-found and fall through to `CreateUser`:

- `server/internal/auth/handlers.go:97` (magic link verify)
- `server/internal/auth/handlers.go:146` (dev session)
- `server/internal/auth/google.go:101` (Google OAuth)
- `server/cmd/devseed/main.go:75` (dev seed)

If the DB is unreachable or the query is cancelled, these paths will attempt an insert that also fails, masking the real cause.

## Proposal

Add a sentinel:

```go
var ErrUserNotFound = errors.New("user not found")
```

Return it from all three `GetUserBy*` methods, and update the four call sites to use `errors.Is(err, auth.ErrUserNotFound)` before falling through to `CreateUser`. Genuine DB errors then surface directly.

## Scope

- Auth package change: tiny, mechanical.
- Four call sites need a one-line tweak each.
- No behavior change on the happy path; better error reporting when the DB misbehaves.

justinlindh:
Closes #207.

## Summary

- Add exported `auth.ErrUserNotFound` sentinel returned by `GetUserByEmail`, `GetUserByGoogleID`, and `GetUserByID`.
- Update the four "find-or-create" call sites to fall through to `CreateUser` only on `errors.Is(err, auth.ErrUserNotFound)`; surface other errors (connection loss, context cancellation, etc.) with an explicit 500 or wrapped error.
  - `server/internal/auth/handlers.go` -- magic link verify
  - `server/internal/auth/handlers.go` -- dev session
  - `server/internal/auth/google.go` -- OAuth callback (both the Google-ID and fallback email lookups)
  - `server/cmd/devseed/main.go` -- dev seed utility
- Switch the `sql.ErrNoRows` equality checks to `errors.Is` while I was in the file.
- Add not-found integration tests for all three lookup methods.

Happy path is unchanged; the fix changes only what happens when a lookup returns an error other than `sql.ErrNoRows`.

## Test plan

- [x] `go vet ./...`
- [x] `go test ./...` (unit) passes
- [x] `go test -tags=integration ./internal/auth/ ./cmd/devseed/` passes against a fresh Postgres, including the three new `_NotFound` tests
- [x] Full integration suite passes with `-p=1` (serial) on a clean DB


justinlindh:
@claude

justinlindh:
Nit addressed in 7728835: `ValidateSession` and `ValidateMagicLink` now use `errors.Is(err, sql.ErrNoRows)`, matching the `GetUserBy*` methods in the same file.

On the middleware observation: agreed, it is a separate concern (no find-or-create semantics, and pre-existing behavior). Leaving it for a follow-up rather than expanding this PR.

Issue #214, opened 2026-04-22
ui(dialup): port intercom redesign decisions to dialup theme


justinlindh:
## Scope

After PR #218 (intercom theme redesign) merges, port equivalent changes to the dialup theme. Both themes need to stay aligned.

## 1:1 copy changes (direct translation, same words)

- "Line name" => "Handset name" (placeholder: `Kitchen, Grandma's bedroom, Garage`, helper: "Most families name handsets by where they live")
- "Pair a device" => "Pair a new handset"
- "Generate invite code" => "Invite a friend"
- "Call history" (page title and rail item) => "Call log"

## Structural changes (reskinned, same layout)

- **Dashboard**: replace KPI-strip-heavy layout with per-line status cards that include a "Talking to <name>" active-call callout. Dialup treatment: beveled 3D status panels with a blinking ON-AIR-style badge for active calls.
- **Families**: replace compact list with two-pane address-book cards (identity + connected-since on left, their lines on right). Keep the hub-and-spoke hero above. Dialup treatment: Win98-style address book chrome with the hub as a network-neighborhood graphic.
- **Settings**: two-column sticky side-nav + theme cards with actual swatches (intercom palette: `#f5f1ea, #2e231b, #c48b3a`; dialup palette: `#0a3a8a, #7ab4ff, #f0c020`).

## Rejected in both themes

- "Lines" => "Rooms/Handsets" rename (keep "Lines"; a line can have multiple handsets)
- "Connected families" => "Your neighborhood"
- "Status: online/offline" => "Ready / asleep"
- "Accept invite" => "Enter someone's code"

## Cross-theme features tracked separately (do dialup treatment when each ships)

- Firmware changelog: #211
- Multi-handset per line: #212
- /call/live observation deck: #213

## Related
- Design bundle for the intercom review lives on branch `fix/auth-err-user-not-found` worktree

Issue #213, opened 2026-04-22
feat(server): /call/live observation deck with link-health telemetry


justinlindh:
## Problem

When a line is on a call, the dashboard shows "Active calls: 1" and clicking it does nothing useful. The live session is the most interesting moment in the system, and right now it is a counter.

## Proposal

A dedicated `/call/live` route reachable from any lit-up status chip, rendering the signaling-layer view of an in-progress call.

### Content
- Both endpoints as rolodex cards (line name, number, household)
- Duration ticker updated via SSE or htmx polling
- Link-health bars sourced from RTCP receiver reports (loss, jitter)
- Connection type, RTT, packet loss percentages
- "Force disconnect (signaling)" as a signaling-layer teardown; clearly labeled as a session-termination action, not media access

### Non-goals
- Anything that could read as audio access or waveform
- Any copy that ambiguously suggests the server sees call content

### Copy constraints
All on-page copy must carry the disclaimer that the server sees the pipe, not the audio. A representative line: "Bars show packet loss and jitter reported by each endpoint. We never see the audio itself."

## Prerequisites
- RTCP stats forwarded from peers to signald (not collected today)
- Signaling command to tear down an in-progress session (new capability)
- Update mechanism for the page (SSE preferred; htmx polling acceptable)

## Theme UX
Both intercom and dialup. Intercom leans on the rolodex/paper metaphor. Dialup can lean into "MODEM CONNECTED" chrome. Do intercom first; schedule dialup after.

## Related
- Intercom redesign mockup exists in the design bundle (Call live artboard)

Issue #212, opened 2026-04-22
feat(server): UX for multiple handsets on a single line


justinlindh:
## Problem

A "line" is a phone number; a "handset" is a physical paired device. The current UI conflates them: each pair action creates a new line, and there is no way to attach a second handset to an existing line.

The party-line feature (on the `worktree-party-line` branch; not yet merged) introduces multiple handsets ringing one line. The web UI needs to catch up before that branch lands, or immediately after.

### Pair flow
Add a choice on `/phones` pair form:
- **Add to existing line** (dropdown of current lines, defaults to off)
- **Create a new line** (default)

### Lines listing
Today: one row per line, one implicit handset.
New: one row per line; when more than one handset is paired, show a nested handset list (Kitchen, Porch) under the line row.

### Identity split
- Line owns: phone number, routing
- Handset owns: name, firmware version, online status

## Dependencies
- Merge of `worktree-party-line` branch
- `docs/architecture/party-line.md` (on that branch) describes the state machine and data model

## Out of scope
- Unpairing individual handsets from a multi-handset line (separate issue when needed)
- Mixed-firmware warnings across handsets on one line

## Theme UX
Both intercom and dialup.

## Related
- Called out in the intercom redesign (Lines critique)

Issue #211, opened 2026-04-22
feat(server): firmware changelog surface for handset updates


justinlindh:
## Problem

When a handset firmware update is available, users have no way to see what changed. The server tracks firmware versions per handset but the UI does not surface release notes, so users cannot decide whether to care about an update.

## Proposal

Track per-version release notes server-side and show them wherever a handset is referenced.

### Surfaces
- Lines row: inline chip `fw 1.3 · update available` that expands to release notes on click
- Pair flow: confirmation text ("paired on v1.4") after pair succeeds
- Optional Dashboard summary card when one or more handsets have updates pending

### Data
- `firmware_releases` table: `version`, `released_at`, `notes_md`, `critical` (reserved for security fixes)
- Notes authored in the firmware release commit or a tracked markdown manifest; server imports at startup

## Out of scope
- Auto-update flow (user-initiated only for now)
- Per-handset update scheduling

## Theme UX
Both intercom and dialup. Intercom uses inline chip + expandable panel; dialup uses a "what's new" MOTD banner on first visit after an update ships.

## Related
- Surfaced during the intercom redesign (Lines V3 empty-first mockup)

PR #218, opened 2026-04-22
feat(server): intercom theme redesign

justinlindh:
## What

Applies the Claude Design review bundle to the intercom theme of the Digits web app. Dashboard becomes a room-card grid with active-call callouts, pair flow nudges users toward room-style handset names, Settings gets a sticky side-nav + theme swatch previews, Families swaps the compact list for a two-pane Neighborhood layout, and the invite code renders as a postcard.

## Why

The dashboard was three KPIs of zero plus a duplicate of the Lines page; the Lines pair field had no vocabulary nudge; Families leaned on names without context; Settings was an unscannable scroll. This PR applies a focused rework that fixes those without renaming the core abstractions (lines stay lines; online/offline stays online/offline).

## Scope

Intercom theme only. Templates are shared across themes, so `dialup.css` ships placeholder rules for every new class to avoid broken rendering during rollout. Full dialup treatment tracked in #214.

## Copy changes

- "Pair a device" -> "Pair a new handset"
- "Line name" -> "Handset name" (placeholder: `Kitchen, Grandma's bedroom, Garage`; helper: "Most families name handsets by where they live.")
- "Generate invite code" -> "Invite a friend"
- "Call history" -> "Call log"

Rejected: `Lines` stays (see #212), `Connected families` stays, online/offline stays, Accept invite stays.

## Structural changes

- **Dashboard**: KPI strip removed. Rooms grid with active-call callout (`Talking to <peer> · mm:ss elapsed`). Conditional Today panel (only when call history is enabled) with `N calls · M min total` sub-label. Connected-families chip row.
- **Lines**: renamed field + Firmware column (placeholder `unknown` until #211).
- **Families**: two-pane Neighborhood replaces compact list; hub-and-spoke hero kept; invite code shown as postcard with `To our neighbors,` greeting + spec-locked body copy; multiple pending invites allowed with per-row Cancel.
- **Settings**: two-column sticky nav, theme cards with palette swatches, strengthened Privacy copy.
- **Confirm dialog**: page-level `<dialog>` pattern (data-confirm-* attributes) replaces browser `confirm()` on Cancel pending invite and Disconnect family. Distinct banners: `Invite canceled.` vs `Family disconnected.`
- **Sticky rail**: top nav bar stays pinned on scroll.

### Dashboard

Dashboard

### Lines

Lines

### Families - invite postcard (with multiple pending invites allowed)

Families postcard

### Families - default state

Families default

### Settings

Settings

### Call log

Call log

### Themed confirm dialog

Confirm dialog

## Spec-compliance automation

Every screen in the redesign now has an integration test in `server/internal/web/spec_compliance_test.go` that asserts the rendered HTML contains the exact strings, classes, and structural hooks the spec requires. One test per screen (`TestSpecDashboard`, `TestSpecLines`, `TestSpecFamilies`, `TestSpecSettings`, `TestSpecCallLog`), each assertion inline-commented with the spec line it enforces. Run:

```
go test -tags=integration -run 'TestSpec' ./internal/web/...
```

File top-comment points back at the spec doc and documents the update discipline: when the spec moves on, update the spec file AND the matching assertion in the same commit.

## Follow-ups (separate issues)

- #211 firmware changelog (wire real data into the Firmware column)
- #212 multi-handset per line (party-line UX)
- #213 `/call/live` observation deck (RTCP-backed)
- #214 dialup theme port (full CSS treatment of all new classes)

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./...` passes (unit tier)
- [x] `go test -tags=integration ./internal/web/...` passes (integration tier, incl. new spec-compliance suite)
- [x] Manual smoke in intercom theme via `make dev-up` with seeded lines + linked family (screenshots above)
- [ ] Manual smoke in dialup theme: no broken layouts on any surface


justinlindh:
@claude

justinlindh:
Addressed all three bugs plus the copy tweak in 54ea3cd:

- **Dashboard Today panel household scoping**: now uses RecentForPhones with the household's own line numbers.
- **Today panel timezone boundary**: now anchors midnight on the household's configured timezone via householdContext's loc, and renders StartedAt in that zone too.
- **phone-edit-row colspan**: bumped 3 to 4 so the inline edit form spans Label + Status + Firmware + Added after the new column.
- **Postcard greeting**: 'To our neighbors,' -> 'To our friends,' per product direction.

Also updated the Playwright e2e assertions that were still checking the pre-redesign copy/structure (the 5 failures in the Server E2E job).

Non-blocking observations acknowledged, not acting on:
- **'0 min total' rounding for <=30 sec of calls**: behavior is technically correct nearest-rounding; no change.
- **data-confirm-action trust boundary**: current values are server-rendered UUIDs; noting it for future interpolation of user-controlled data.
- **Spec-test shared test@example.com user**: today's integration runner is sequential, so not a flake source now; noted if we ever go parallel.

PR #222, opened 2026-04-22
feat(server): per-phone silent mode toggle (#217)

### LINKED ISSUE #217: feat: add option to silence ringer from web app per device
It would be desirable to be able to place the phones in "silence" mode (quiet hours kind of thing). Let's implement a feature where this can be set via the web app (both themes). State will be set on the device and usual state sync requirements apply.

Effectively, "quiet mode" will simply disable the ringer. The LED should still flicker and calls can be answered and placed. The ringer just gets disabled (no bell).

justinlindh:
## What

Adds a per-phone Silent mode toggle to the web app. When on, incoming calls suppress the bell on that phone but still flash the LED, and offhook still answers. Setting persists server-side and locally on the device, syncs over the existing signaling channel, and survives reboots and disconnects.

## Why

Closes #217. Quiet hours / nighttime scenarios where you want a phone to stop ringing without losing call functionality.

## Key choices

- Silent flag rides on the existing `LineSettings` JSONB blob and `TypeLineSettings` wire message. No schema migration. Mirrors the existing `voice_style` plumbing end to end.
- Gate sits in digitsd's phone controller (`onSignalRing`): skip `SendRing(true)` when silent, still send `SendLED("BLINK")`. Firmware is untouched.
- Asymmetric mid-ring behavior: flipping silent ON during an active ring stops the bell immediately. Flipping OFF mid-ring does NOT start it. (A bell starting mid-ring would be jarring; if you want it, just answer.)
- Same HTML markup in both themes; `digits.css` and `dialup.css` carry all the visual differentiation.

## Screenshots

**Intercom theme**

Detail page with the Ringer panel (silent on):

intercom detail

Lines list with the muted-bell glyph next to Kitchen:

intercom list

**Dialup theme**

Detail page with the Win95 checkbox in the Ringer groupbox:

dialup detail

Lines list with the amber answering-machine LED + SILENT label:

dialup list

## Test plan

- [x] Unit: `server/internal/line/...` round-trips the bool through JSONB
- [x] Unit: `pi/digitsd/internal/config/...` persists and reloads `silent_mode` from disk
- [x] Unit: `pi/digitsd/internal/phone/...` covers the gate in `onSignalRing`, the mid-ring ON interrupt, the asymmetric mid-ring OFF no-op, idle side-effect-freedom
- [x] Unit: `server/internal/web/handler_test.go` covers form parsing (on / off / missing), 404 on unknown line, push-to-connected-device, no-op skip, and the htmx partial response
- [x] Integration: `server/internal/signaling/relay_linesettings_test.go` verifies the register-time push carries the new flag in both true and false cases
- [x] E2E: `server/e2e/tests/09-silent-mode.spec.ts` exercises the toggle round-trip and badge appear/disappear in both themes (skips cleanly without auth or phones)
- [ ] Hardware: deploy to D1 and walk the seven scenarios (baseline ring, silent ON suppresses bell, mid-ring ON stops bell, mid-ring OFF does not start bell, reboot persists, offline toggle applies on reconnect, `config.json` reflects state). Pending; will run before merge.

## Notes for reviewers

- A pre-existing redundancy in `pushLineSettings` (double `hub.Get` + TOCTOU window) was found by the simplify pass and fixed in this branch. Introduces `signaling.ErrNotConnected` as a typed sentinel.
- The `.toggle` CSS class collision with the existing privacy pill-switch on `/settings` was caught in review and fixed by renaming to `.checkbox-row`.
- The mobile card grid bug (badge becoming a 5th grid child and breaking the row) was caught in review and fixed by nesting the badge inside `.lines__name`.
- After merging main, `handlePhoneSilentModePost` was aligned with PR #220's `requireLineOwnership` pattern so the new endpoint enforces household scoping.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -490,6 +490,29 @@ func (d *daemonCallbacks) setVoiceStyleConfig(style string) error {
 	return d.cfg.Save()
 }
 
+// applySilentModeLive forwards a silent-mode change to the phone controller.
+// Safe before the controller is constructed: becomes a no-op, and
+// setSilentModeConfig still persists the flag for the next startup.
+func (d *daemonCallbacks) applySilentModeLive(silent bool) {
+	d.mu.Lock()
+	ctrl := d.ctrl
+	d.mu.Unlock()
+	if ctrl == nil {
+		return
+	}
+	ctrl.SetSilentMode(silent)
+}
+
+// setSilentModeConfig writes the new flag to the local config cache under
+// d.mu (serializing with call-setup paths that read d.cfg) and then persists
+// it to disk outside the lock.
+func (d *daemonCallbacks) setSilentModeConfig(silent bool) error {
+	d.mu.Lock()
+	d.cfg.SilentMode = silent
+	d.mu.Unlock()
+	return d.cfg.Save()
+}
+
 // triggerHangup dispatches a hangup to the controller from a fresh goroutine.
 // Callers holding d.mu must use this to avoid deadlock: HandleSignal calls
 // back into HangupCall, which also acquires d.mu.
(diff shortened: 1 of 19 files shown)


justinlindh:
@claude

PR #230, opened 2026-04-22
fix(server): move update <details> out of lines__row anchor on mobile

justinlindh:
## What

On the mobile Lines card, the update chip was rendering as a bare "▶ Details" fallback marker below each row instead of an "update" chip with notes inside. Moves the `<details>` element out of the surrounding `<a>` anchor so the HTML parser stops doing error-recovery.

## Why

HTML spec forbids interactive elements (`<details>` is one) inside `<a>`. Browsers recover by aborting and restarting the `<a>` around the invalid child. That breaks rendering in two ways:

- `<summary>` content does not paint; the browser shows the default triangle + "Details" text.
- The original `<a>` tree gets fragmented. DOM inspection showed `.lines__row` inflating from 3 actual elements to 15 matches.

The desktop table is unaffected: it uses `<tr>` / `<td>`, not `<a>`, so the same `<details>` structure was always valid there.

## Reproduction (prod, forced mobile layout)

Before: ▶ Details fallback text below each row

## Fix

Restructure the mobile card so `<details>` is a sibling of the `<a>`:

```html
<li>
  <a class="lines__row" href="/phones/{number}">
    ...online/offline inline...
  </a>
  {{if update notes}}
    <details class="update-details update-details--mobile">
      <summary class="chip chip--warn">update available</summary>
      <div class="update-notes">...</div>
    </details>
  {{end}}
</li>
```

- Online/offline returns to being shown inline inside the row (matches pre-update-chip behavior for devices with no update).
- When an update is available, a full-width chip sits below the row and expands into the stacked per-version groomed notes.
- Added a `.update-details--mobile` modifier in both themes so the sibling block fills the card width with sensible spacing.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./...` green
- [x] `go test -tags=integration ./internal/web/` green
- [x] DOM inspection confirms `<details>` is now a `<li>` child, not inside `<a>`. Parser will not fragment the anchor.
- [ ] After-merge deploy: take after screenshot from prod at mobile viewport to confirm chip renders correctly.

### DIFF
--- a/server/internal/web/static/dialup.css
+++ b/server/internal/web/static/dialup.css
@@ -2365,3 +2365,8 @@ body.dialup .phone-silent__text {
   padding: 0 4px;
 }
 .pair-banner__close:hover { color: #ffffff; }
+
+/* Mobile variant: full-width details sibling below the .lines__row anchor. */
+.update-details--mobile { display: block; margin: 0 10px 8px; }
+.update-details--mobile > summary { display: inline-flex; align-items: center; gap: 4px; }
+.update-details--mobile .update-notes { max-width: none; margin-top: 6px; }
--- a/server/internal/web/static/digits.css
+++ b/server/internal/web/static/digits.css
@@ -1638,3 +1638,8 @@ details.disclosure > *:not(summary) { border-top: 1px solid var(--rule-soft); }
   padding: 0 4px;
 }
 .pair-banner__close:hover { color: #2e231b; }
+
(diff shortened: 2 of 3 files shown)


justinlindh:
Merged + deployed. "After" shot from prod with mobile card layout forced visible:

After: update available chip renders correctly below each row

DOM query now returns 3 `.lines__row` matches for 3 lines (down from 15 caused by the parser tree-fixup). Chip renders, expansion works, online/offline inline as expected.

PR #231, opened 2026-04-22
Potential fix for code scanning alert no. 5: Cookie 'Secure' attribute is not set to true

justinlindh:
Potential fix for [https://github.com/justinlindh/digits/security/code-scanning/5](https://github.com/justinlindh/digits/security/code-scanning/5)

Set `Secure: true` on the logout `admin_session` cookie in `server/internal/admin/server.go` (inside `handleLogout`), matching the login cookie settings.

Best single fix without changing functionality:
- In the `http.Cookie` literal passed to `http.SetCookie` in `handleLogout`, add:
  - `Secure:   true,`
- Keep existing fields unchanged (`Name`, `Value`, `Path`, `HttpOnly`, `MaxAge`).
- No new imports, methods, or dependencies are required.

This ensures all writes to `admin_session` enforce HTTPS-only transmission.

### DIFF
--- a/server/internal/admin/server.go
+++ b/server/internal/admin/server.go
@@ -178,6 +178,7 @@ func (s *Server) handleLogout(w http.ResponseWriter, r *http.Request) {
 		Value:    "",
 		Path:     "/admin",
 		HttpOnly: true,
+		Secure:   true,
 		MaxAge:   -1,
 	})
 	http.Redirect(w, r, "/admin/login", http.StatusSeeOther)


justinlindh (commented):
The login cookie at line 163 already sets both HttpOnly and Secure. The logout cookie at line 176 had HttpOnly but was missing Secure. This fix makes the two cookies consistent. admind is only served over HTTPS in production, so Secure on a clearing cookie is correct and harmless. LGTM.

PR #232, opened 2026-04-22
Potential fix for code scanning alert no. 4: Email content injection

justinlindh:
Potential fix for [https://github.com/justinlindh/digits/security/code-scanning/4](https://github.com/justinlindh/digits/security/code-scanning/4)

General fix: never place untrusted data directly into raw email headers. Validate recipient address format and reject any header value containing CR (`\r`) or LF (`\n`) before building the SMTP message. Keep body handling separate from headers.

Best targeted fix here (without changing intended behavior): in `server/internal/email/email.go`, add strict validation in `SMTPSender.Send`:
1. Reject `to`, `subject`, and `s.from` if they contain CR/LF (prevents header injection).
2. Parse/validate `to` and `s.from` as email addresses using `net/mail`.
3. Use parsed canonical addresses in headers and SMTP envelope recipient list.
4. Return an error on invalid input rather than sending malformed/injectable mail.

This requires adding imports: `errors` and `net/mail`.

### DIFF
--- a/server/internal/email/email.go
+++ b/server/internal/email/email.go
@@ -1,6 +1,8 @@
 package email
 
 import (
+	"errors"
+	"net/mail"
 	"net/smtp"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


justinlindh (commented):
The fix is correct and thorough. CRLF injection in To/Subject/From is blocked up front. net/mail.ParseAddress is used to validate and normalize both addresses before building headers, and fromAddr.Address / toAddr.Address (the bare address without display name) are passed to smtp.SendMail so the envelope is clean. One note: subject is not passed through ParseAddress, so a CRLF in subject is caught by the ContainsAny guard but not normalized. That is fine because subjects are not email addresses and net/mail does not parse them. The guard covers the case. LGTM.

PR #233, opened 2026-04-22
Potential fix for code scanning alert no. 9: Disabled TLS certificate check

justinlindh:
Potential fix for [https://github.com/justinlindh/digits/security/code-scanning/9](https://github.com/justinlindh/digits/security/code-scanning/9)

To fix this safely without changing intended behavior, remove the explicit insecure TLS override so Go’s default TLS verification is used. In `pi/digitsd/cmd/latclient/main.go`, replace the `websocket.Dialer` initialization that sets `InsecureSkipVerify: true` with a default dialer configuration (`websocket.Dialer{}`), allowing standard certificate and hostname validation.

Concretely:
- Edit the dialer block around lines 45–47.
- Remove dependency on `crypto/tls` in this file since it is no longer used.
- No new methods/definitions are required; no behavior changes besides enforcing proper TLS verification.

### DIFF
--- a/pi/digitsd/cmd/latclient/main.go
+++ b/pi/digitsd/cmd/latclient/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"crypto/tls"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -42,9 +41,7 @@ func main() {
 	q.Set("number", *number)
 	u.RawQuery = q.Encode()
 
-	dialer := websocket.Dialer{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-	}
+	dialer := websocket.Dialer{}
 	ws, _, err := dialer.Dial(u.String(), nil)
 	if err != nil {
 		log.Fatalf("dial signald: %v", err)


justinlindh (commented):
Do not merge as-is. latclient is a dev/lab tool: the default URL flag is wss://localhost:8443/ws and the tool SSHs into a Pi by IP to inject a hook event. Its only known use case is connecting to a local signald instance that uses a self-signed TLS cert. Removing InsecureSkipVerify unconditionally makes the tool fail in that scenario. The right fix is to add an -insecure flag (default false) that enables InsecureSkipVerify when set, so the CodeQL alert is satisfied in the default path while the dev use case still works. Dropping the flag entirely is too blunt.

justinlindh:
Closing: the review (#233) flagged that `latclient` defaults to `wss://localhost:8443/ws` against a self-signed local signald, so removing `InsecureSkipVerify` unconditionally would break the only known use case. Safer path is a dedicated `-insecure` flag (defaulting to `false`) so the CodeQL alert resolves on the default path without breaking the dev workflow. Happy to reopen a replacement PR with that shape.

PR #234, opened 2026-04-22
Potential fix for code scanning alert no. 10: Cookie 'HttpOnly' attribute is not set to true

justinlindh:
Potential fix for [https://github.com/justinlindh/digits/security/code-scanning/10](https://github.com/justinlindh/digits/security/code-scanning/10)

Set `HttpOnly: true` on both cookie literals inside `clearSessionCookie` in `server/internal/auth/middleware.go`.

Best minimal fix without changing behavior:
- In the first `http.Cookie{...}` (domain-scoped delete cookie), add `HttpOnly: true`.
- In the second `http.Cookie{...}` inside `if domain != ""` (host-scoped delete cookie), add `HttpOnly: true`.

This preserves existing logic (same name/domain/path/max-age deletion behavior) while ensuring all emitted `digits_session` cookies are marked HttpOnly, addressing both alert variants at once.

### DIFF
--- a/server/internal/auth/middleware.go
+++ b/server/internal/auth/middleware.go
@@ -59,16 +59,18 @@ func UserFromContext(ctx context.Context) *User {
 // Writing both variants mops up either form.
 func clearSessionCookie(w http.ResponseWriter, domain string) {
 	http.SetCookie(w, &http.Cookie{
-		Name:   CookieName,
-		Domain: domain,
-		MaxAge: -1,
-		Path:   "/",
+		Name:     CookieName,
+		Domain:   domain,
+		MaxAge:   -1,
+		Path:     "/",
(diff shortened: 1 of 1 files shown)


justinlindh (commented):
The fix is cosmetically correct but note that #235 also touches both cookie literals in clearSessionCookie to add Secure: true, and the two PRs conflict. They cannot both be merged cleanly; only one will apply without a merge conflict. The preferred approach is to merge them together into a single PR that adds both HttpOnly and Secure to both clearing cookies. The login cookie in handlers.go already has both flags, so both are needed on the clear path for consistency. If merging separately, apply #235 first since Secure is the more security-relevant attribute, then rebase this PR on top.

PR #235, opened 2026-04-22
Potential fix for code scanning alert no. 8: Cookie 'Secure' attribute is not set to true

justinlindh:
Potential fix for [https://github.com/justinlindh/digits/security/code-scanning/8](https://github.com/justinlindh/digits/security/code-scanning/8)

Set `Secure: true` explicitly on every `http.Cookie` in `clearSessionCookie` so cookie-clearing headers are only sent as secure cookies.

Best minimal fix in `server/internal/auth/middleware.go`:
- In `clearSessionCookie`, update both `http.Cookie` literals (the domain-scoped one and the host-scoped one) to include `Secure: true`.
- Keep all existing behavior unchanged (`Name`, `Domain`, `MaxAge`, `Path`), only add the missing secure attribute.

No new imports, methods, or dependencies are required.

### DIFF
--- a/server/internal/auth/middleware.go
+++ b/server/internal/auth/middleware.go
@@ -63,12 +63,14 @@ func clearSessionCookie(w http.ResponseWriter, domain string) {
 		Domain: domain,
 		MaxAge: -1,
 		Path:   "/",
+		Secure: true,
 	})
 	if domain != "" {
 		http.SetCookie(w, &http.Cookie{
 			Name:   CookieName,
 			MaxAge: -1,
 			Path:   "/",
+			Secure: true,
 		})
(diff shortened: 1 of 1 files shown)


justinlindh (commented):
The fix is correct. The login cookie at handlers.go:128 and the Google OAuth callback at google.go:151 both set Secure: true. The clearing cookies in clearSessionCookie were missing it. Note that #234 also modifies the same two cookie literals in clearSessionCookie to add HttpOnly: true. The two PRs will conflict and cannot both merge cleanly. Prefer combining them into a single PR that adds both attributes at once, or merge this one first and rebase #234 on top.

PR #236, opened 2026-04-22
Potential fix for code scanning alert no. 1: Workflow does not contain permissions

justinlindh:
Potential fix for [https://github.com/justinlindh/digits/security/code-scanning/1](https://github.com/justinlindh/digits/security/code-scanning/1)

Add an explicit `permissions` block to the workflow so `GITHUB_TOKEN` is least-privileged by default.  
Best fix here: set workflow-level permissions to `contents: read`, since the only action requiring token access is `actions/checkout`, and build/test/vet steps do not need write scopes.

Edit **`.github/workflows/server-ci.yml`** near the top-level keys (after `on:` block is a clear location) and add:

- `permissions:`
  - `contents: read`

No imports, methods, or additional definitions are needed.

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -5,6 +5,9 @@ on:
     branches: [main, 'feature/**']
   pull_request:
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}


justinlindh (commented):
The workflow only checks out code, builds the binary, runs unit tests, and vets. It does not post PR comments, upload artifacts, push tags, or write to any GitHub resource. contents: read is the correct least-privilege default and does not restrict anything this job actually does. LGTM.

PR #252, opened 2026-04-22
feat(server): link call log entries to observation deck

justinlindh:
## What

Adds a `Details →` link to each non-conference row on `/calls`, pointing at `/call/live/{id}`. Covers both the desktop table (new trailing column) and the mobile card list. Conference rows stay unlinked since the deck is strictly two-party.

## Why

The observation deck already supports ended calls via DB readback (`handleCallLiveDetail` explicitly does not 404 on `status='ended'`), but there was no UI entry point — users had to hand-type the URL. This surfaces postmortem jitter/loss/bytes data for any ended call from the normal call log view.

Note: this only adds the link for 2-party calls. Live metrics for 3-way conferences are not implemented; the relay's `handleLinkHealth` drops samples for conference members because `CallIDFor` only scans the 2-party active map, and `call_link_health` has no conference_id key. Conference rows remain unlinked until that plumbing lands.

## Test plan

- [x] `go test ./...` (unit) green in worktree
- [x] `make test-integration` green (all integration tests including `TestCallsPageScopedToHousehold`, `TestCallsPageRenders3WayConference`, `TestCallsPageReturns200` — all exercise the real template render)
- [x] `golangci-lint run ./...` — 0 issues
- [ ] Browser visual check in both intercom and dialup themes — not performed; port 8080 was occupied and the change is a single-column table addition + one mobile anchor. Happy to screenshot if you want.

### DIFF
--- a/server/internal/web/templates/calls.html
+++ b/server/internal/web/templates/calls.html
@@ -31,6 +31,7 @@
             <th>Status</th>
             <th>Started</th>
             <th>Duration</th>
+            <th class="r"></th>
           </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

PR #267, opened 2026-04-23
fix(server): render saved timezone server-side in settings

justinlindh:
## Summary

- Timezone dropdown on `/settings` now renders its `selected` option server-side, matching the pattern every other setting already uses (theme, CRT mode, call-history toggle, voice style).
- Removes the inline script that silently replaced the selection with `Intl.DateTimeFormat().resolvedOptions().timeZone` whenever the saved value was `UTC`. That override caused two user-visible failure modes both reported as "timezone does not persist":
  1. Users who explicitly picked UTC saw their browser zone on reload and believed their selection was lost.
  2. Fresh households (DB default `UTC`) saw the browser zone pre-selected, assumed it was saved, and never clicked Save. Timestamps across the app continued rendering in UTC via `Household.Location()` (dashboard, calls, phones), which looked broken.
- Adds `TestSettingsPageRendersSavedTimezone`: sets a household to `America/Chicago`, GETs `/settings`, asserts the body contains `value="America/Chicago" selected` and that no other zone carries `selected`. This would have caught the original bug.

Server-side persistence itself (handler, store, migration) was already correct and covered by `TestSettingsTimezonePost`. The bug lived entirely in the template's client-side "helpful" detection.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./...` (0 issues)
- [x] `make test`
- [x] `make test-integration` (new render test passes)
- [ ] Manual: load `/settings`, pick UTC, save, reload; dropdown still shows UTC.
- [ ] Manual: load `/settings` as a brand-new household; dropdown shows UTC (honest default) instead of pretending the browser zone was saved.

### DIFF
--- a/server/internal/web/handler_test.go
+++ b/server/internal/web/handler_test.go
@@ -246,6 +246,40 @@ func TestSettingsTimezonePost(t *testing.T) {
 	}
 }
 
+func TestSettingsPageRendersSavedTimezone(t *testing.T) {
+	h, database, authStore := setupHandler(t)
+	cookie, hh := setupAuthedHousehold(t, h, database, authStore)
+
+	if err := h.householdStore.SetTimezone(context.Background(), hh.ID, "America/Chicago"); err != nil {
+		t.Fatalf("SetTimezone: %v", err)
+	}
+
+	req := httptest.NewRequest(http.MethodGet, "/settings", nil)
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

Issue #280, opened 2026-04-24
Misc fixes: web handler hygiene + template em dashes


justinlindh:
## Web handler hygiene

Three related problems in `server/internal/web/`. Worth doing together; #2 unblocks #3.

### 1. N+1 `GetForUser` per protected page render

Several handlers issue redundant `householdStore.GetForUser` calls because each chrome value (name, call-history flag, DND flag, timezone) is fetched independently:

- `handleDashboard`: ~4 calls per render (one in `householdContext`, one for the `householdID` lookup, two more inside `buildLinesData`).
- `handleLinksGet`, `handleCalls`, `handleSettings`: 2 each.

Fix: thread a single `*household.Household` through the handler chain (or cache it on the request context once `householdContext` resolves it). Saves 1-3 round-trips per page render.

### 2. Page-data struct sprawl

Nine page-data structs (`dashboardData`, `linesData`, `lineDetailData`, `callsData`, `linksData`, `settingsData`, `onboardData`, `conferenceLiveDetailData`, `connectingData`, `callLiveDetailData`) each carry the same chrome quintuplet: `Page`, `Version`, `CallHistoryEnabled`, `HouseholdName`, `HouseholdDND`. Every new chrome field touches all 9 files (the DND chip rollout was a recent example).

Fix: introduce a `chromeData` embedded struct populated by one helper; embed it in each page-data struct. Future chrome additions become a 1-line change.

### 3. `settingsData` carries fields already on its embedded `Household`

`settingsData` already embeds `Household *household.Household` but also stores `HouseholdName`, `HouseholdDND`, and `CallHistoryEnabled` separately, then hand-copies them in the handler. Drop the derived fields; have the template read `.Household.Name` / `.Household.DoNotDisturb` / `.Household.CallHistoryEnabled` directly.

This depends on #2 to stay consistent (other pages don't currently embed `*Household`).

### 4. Pre-existing em dashes violate the no-em-dash hard rule

The pre-commit hook only blocks new content, so these have lived past it:

- `templates/layout-v2.html:6` (`<title>`)
- `templates/layout-dialup.html:6` (`<title>`)
- `templates/layout-dialup.html:22` (brand span)
- `templates/phone-detail.html:250, 256, 359`
- `templates/call-live-detail.html:1` (`<title>`)

Replace each with a colon, period, comma, or restructure per the project hard rule.

PR #281, opened 2026-04-24
feat(server): household-wide do-not-disturb toggle (#269)

### LINKED ISSUE #269: feat(server): house-wide "do not disturb" master switch
**Candidate feature, not necessarily planned.** Filing so it's not lost.

Per-device ringer silence shipped in #217 (closed). A household-level DND master switch would silence every paired handset at once with a single action from the web UI, for the common cases where the household wants one toggle rather than four: "everyone's asleep," "we're out," etc.

Behavior sketch:
- Master toggle at the household level, not per line.
- When on, every paired handset's ringer is silenced using the same semantics as #217 (LED still flickers, calls can still be answered and placed, just no bell).
- Synced to each digitsd via the same state-sync mechanism as per-device silence.
- Composes with per-device silence: the handset rings only if neither household DND nor its own silence flag is set.

Open questions:
- Persisted across restarts, or session-only?
- Dashboard banner while active so it's not forgotten on?
- Does it also apply to incoming calls via linked families, or only local ring?

justinlindh:
## What

Adds a household-wide Do Not Disturb master switch to the web app. One toggle in Settings silences every paired handset's bell at once, composing additively with the per-line `silent_mode` flag from #217 / #222 (a phone rings only when both household DND and its own silent flag are off). A persistent chip in the top nav surfaces "DND on" across every protected page so it doesn't get left on by accident.

## Why

Closes #269. The "everyone's asleep" / "we're out" use case where a single tap beats hitting four per-device switches. Filed misc-cleanup follow-ups separately as #280.

## Key choices

- New `do_not_disturb BOOLEAN` column on the `households` table (mirrors `call_history_enabled`). No firmware change. No digitsd change. No wire-protocol change.
- Server OR's `household.do_not_disturb` into the existing `LineSettings.SilentMode` field before sending. The wire field's meaning becomes "effective silent" rather than "per-line silent" but the schema, message type, and shape are unchanged. digitsd reads it the same way it does today.
- Single canonical helper `line.EffectiveSilent(s Settings, householdDND bool) bool`. Three push paths (relay reconnect via `OnRegistered`, per-line settings POST, household DND fan-out) all delegate to it.
- New `Store.EffectiveSettingsByNumber` does a single-query JOIN of lines + households so `OnRegistered` doesn't make two round-trips.
- Fan-out on toggle: handler pushes the new effective bool to every connected line in the household. Disconnected devices pick up the right state on reconnect via the existing `OnRegistered` push.
- UI: toggle in `/settings` (Privacy section), persistent `chip chip--err` in the top nav across all pages. Reuses the existing `.chip` component (no bespoke CSS).
- Per-line silent badges on `/phones` are unchanged. They reflect per-device intent only, not effective state, so toggling DND doesn't visually mark all four lines as silent.
- Permissions: any household member can toggle DND, matching the call-history precedent.

## Test plan

- [x] Unit: `internal/household/store_test.go` round-trips `SetDoNotDisturb` through Postgres.
- [x] Unit: `internal/line/settings_test.go` covers the `EffectiveSilent` truth table (4 cases).
- [x] Unit: `internal/web/handler_settings_test.go` table-driven `TestHandleSettingsDoNotDisturbPost` covers persist + fan-out, off direction, and the OR-survives-DND-off case where a per-line silent flag overrides DND being turned off.
- [x] Unit: `internal/web/handler_settings_test.go::NoHouseholdRedirects` covers the no-household redirect path.
- [x] Integration: `internal/signaling/relay_linesettings_test.go` continues to verify the `OnRegistered` push path (truth-table coverage owned by `TestEffectiveSilent`).
- [x] E2E: `e2e/tests/10-do-not-disturb.spec.ts` toggle round-trip + chip appears across `/`, `/phones`, `/links`, `/settings` in both themes; per-line silent badges unchanged.
- [x] `golangci-lint run ./...` clean after every commit.
- [x] `make test-integration` green after every commit.
- [ ] Hardware (manual, on D1+D2 before merge):
  1. Baseline: DND off, both ring on incoming.
  2. DND on: neither rings on incoming (LED still flickers).
  3. Mid-ring DND on: bell stops on both. (Asymmetric: DND off mid-ring does not start the bell, mirroring per-device behavior.)
  4. Linked-family inbound call with DND on: still no ring.
  5. Server restart with DND on: state survives, devices still silent on next ring.
  6. One device offline at toggle time: reconnects, respects DND via `OnRegistered`.

## Notes for reviewers

- The middle five commits (`9db3ce5`, `b15f440`, `2e8e2c4`) are review-driven fix-ups landed during the in-PR review loop, not separate features. The last five commits (`f4928fb` through `a6b27ec`) are a `/simplify` cleanup pass: chip-component reuse, helper widening to drop a redundant query, test consolidation, removing a redundant truth-table test, and one stray comment. Reviewing by squash-diff is fine; commit-by-commit is also coherent.
- `signaling.LineStore.LineSettingsByNumber` was renamed to `EffectiveLineSettings` in `19ed230`. Internal interface; no external consumers.


justinlindh:
@claude

PR #301, opened 2026-04-24
chore(server): remove em dashes from web templates (#280)

justinlindh:
## Summary

Replaces pre-existing em dashes in four template files with colons or restructured copy. The pre-commit hook only blocks *new* em dashes, so these older ones have lived past it. Fixes sub-issue 4 of #280.

Six replacements across four files:

- `layout-v2.html` and `layout-dialup.html` title tags
- `layout-dialup.html` brand span in the title bar
- `phone-detail.html` firmware-SWD note, version-report wait copy, and update-triggered toast
- `call-live-detail.html` title

## Before / after

`layout-v2.html` and `layout-dialup.html` title:

Before: `{{block "title" .}}Digits{{end}} [emdash] Digits`
After: `{{block "title" .}}Digits{{end}} : Digits`

`layout-dialup.html` brand span:

Before: `<span>Digits [emdash] {{if .HouseholdName}}...{{else}}Phone Network{{end}}</span>`
After: `<span>Digits: {{if .HouseholdName}}...{{else}}Phone Network{{end}}</span>`

Verified with `rg` that zero em dashes remain in `server/internal/web/templates/`. `go build ./...` and `make test` both pass.

Refs #280.

### DIFF
--- a/server/internal/web/templates/call-live-detail.html
+++ b/server/internal/web/templates/call-live-detail.html
@@ -1,4 +1,4 @@
-{{define "title"}}Live session — {{.Caller.DisplayName}} &harr; {{.Callee.DisplayName}}{{end}}
+{{define "title"}}Live session: {{.Caller.DisplayName}} &harr; {{.Callee.DisplayName}}{{end}}
 
 {{define "content"}}
 <div class="page deck" data-started-at="{{.Call.StartedAt.UnixMilli}}">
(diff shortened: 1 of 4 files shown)


justinlindh:
Superseded by #304, which covers the same template em-dash cleanup (plus a few additional occurrences in Go comments and error strings in the same files).

PR #304, opened 2026-04-24
fix(server): consolidate page chrome, cut GetForUser N+1, clear em dashes

justinlindh:
## Summary

Addresses #280.

- **chromeData** holds the five fields every protected page-data struct shared (Page, Version, User, Household) with HouseholdName / HouseholdDND / CallHistoryEnabled exposed as methods that read through the embedded `*Household`. All nine page-data structs (dashboard, phones, phone-detail, calls, call-live, links, settings, onboard, conference-live, connecting) embed it, so future chrome fields are a one-line change instead of touching every handler.
- **primaryHousehold(r)** replaces `householdContext` + `callHistoryEnabled`; ownership helpers (`ownedLinesForUser`, `requireCallEndpointOwnership`, `requireConferenceOwnership`, `requireConferenceHostOwnership`) now return `*household.Household` so the household lookup their auth check already runs feeds back to the handler. Net GetForUser per page render:
  - Dashboard: 4 -> 1
  - Calls / Links / Settings: 2 -> 1
  - Phone-detail / call-live / conference-live: 2 -> 1
  - handlePhoneEditPost / handlePhoneDelete: 2 -> 1 (via `requireLineOwnershipWithHousehold`)
- **settingsData** no longer hand-copies HouseholdName / HouseholdDND / CallHistoryEnabled. Templates continue to read `.HouseholdName` (method on chromeData) for layout chrome and `.Household.Name` / `.Household.Timezone` for the edit form, both backed by the same `*Household` pointer.
- **Em dashes** in layout-v2, layout-dialup, phone-detail, call-live-detail are gone. A handful of stale em dashes in Go comments/error strings under server/internal/web/ that the pre-commit hook had grandfathered are also cleaned up in files this refactor touched.

## Test plan

- [x] `go test ./...` (unit)
- [x] `make test-integration` (integration, against Postgres)
- [x] `golangci-lint run ./...` (zero issues)
- [ ] Spot-check rendered chrome on `/`, `/phones`, `/calls`, `/links`, `/settings`, `/phones/{num}` in dev-up

### DIFF
--- a/server/internal/web/handler_api.go
+++ b/server/internal/web/handler_api.go
@@ -86,11 +86,13 @@ func (h *Handler) handleInternalStats(w http.ResponseWriter, r *http.Request) {
 
 
 func (h *Handler) handleAPIStatus(w http.ResponseWriter, r *http.Request) {
-	ld := h.buildLinesData(r, "")
-	nums := h.householdNumbers(r)
+	hh := h.primaryHousehold(r)
+	ld := h.buildLinesData(r, hh, "")
 
+	nums := make(map[string]bool, len(ld.Lines))
 	var onlineCount int
 	for _, row := range ld.Lines {
+		nums[row.Line.Number] = true
 		if row.Online {
 			onlineCount++
 		}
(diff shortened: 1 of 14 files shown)


justinlindh:
@claude

PR #318, opened 2026-04-26
feat(server): play DTMF tones when pairing keypad buttons are pressed

justinlindh:
## Summary
- Copies the existing `pi/tones/dtmf_0.wav`..`dtmf_9.wav` (10 × ~12KB) into the embedded `/static/audio/dtmf/` FS.
- Hooks tone playback into the existing IIFE in `phones.html` so both keypad clicks and physical numeric keydowns trigger the matching DTMF wav.
- Audio elements are created lazily on first press per digit; visitors who never touch the keypad pay no preload bandwidth.
- Adds an embed-FS test that asserts each `dtmf_<n>.wav` serves with WAV magic bytes from production mode, so future static-FS refactors can't silently drop them.

Both intercom/dialup and answering-machine themes pick up the change automatically: both render `.keypad__btn[data-digit]` buttons against the same script block.

## Test plan
- [x] `make test` (server unit tests, including new embed-FS check)
- [x] Browser verification on `/phones` for the answering-machine theme: clicking a digit advances the pin slot and plays the corresponding tone; physical numeric keydown does the same.
- [ ] Reviewer spot-check on intercom and dialup themes after merge


justinlindh:
@claude

PR #326, opened 2026-04-26
fix(digitsd): keep dial tone responsive after volume / audio-test codes

justinlindh:
## Summary
- Volume codes (\`*#*N\`) and the audio-test code (\`*#8378#\`) left the FSM in \`StateIDLE\` while the daemon's callback restarted the dial-tone loop. Subsequent key presses were ignored by \`onKey\`'s IDLE branch, so the resumed dial tone played until the user hung up. Reported symptom: "I enter the key codes to change the volume but then it doesn't silence the dialtone after I keep pressing buttons."
- Distinguishes terminal codes (shutdown, reboot, setup, repair, factory reset, update) from non-terminal ones (volume, audio test). \`ServiceCodeHandler.AddKey\` now returns a \`ServiceCodeResult\` enum and the event loop dispatches to a new \`Controller.ResetToDialtone()\` (DIALTONE) for non-terminal, \`Reset()\` (IDLE) for terminal.

## Test plan
- [x] \`go test ./...\` passes in \`pi/digitsd\`
- [x] \`golangci-lint run ./...\` clean
- [x] \`go vet ./...\` clean
- [x] New \`TestController_ResetToDialtone_NextKeyStopsTone\` covers the exact bug: after \`ResetToDialtone()\`, the next \`KEY:\` event must transition DIALTONE -> DIALING and emit \`SendTone(STOP)\`
- [ ] Bench: pick up handset, dial \`*#*5\`, double-beep plays, dial tone resumes, press any digit -> tone stops and call attempt begins

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -2010,17 +2010,11 @@ func main() {
 				if !easterEggs.AddKey(key) {
 					switch svcCodes.AddKey(key) {
 					case phone.ServiceCodeTerminal:
-						// Daemon is going down (shutdown, reboot, etc.); FSM
-						// state no longer matters. Reset to IDLE.
 						ctrl.Reset()
 						continue // skip forwarding to controller
 					case phone.ServiceCodeNonTerminal:
-						// User is still off-hook (volume, audio test) and the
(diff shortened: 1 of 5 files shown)


justinlindh:
@claude  /simplify

PR #331, opened 2026-04-26
chore(image): bake digitsd's runtime assets at build time

justinlindh:
## Summary

Follow-up to PR #330. Eliminates the digitsd-vs-first-boot rootfs flip race at its root by removing the need for digitsd to do any runtime asset extraction on a fresh-flashed image. The runtime path stays in place for OTA digitsd updates (new version string and fresh hash both miss the marker, so Extract still runs), but on a clean flash there is no version drift between the digitsd binary and its embedded assets, so there is also nothing to extract.

## What was happening

digitsd's `assets.Extractor.Extract` ran on every boot, hashed its embedded FS, compared against `/data/digits/asset-version`, and on mismatch did `sudo mount -o remount,rw / ; cp ... ; remount,ro /`. Fresh flashes never had a marker file, so the rewrite always ran on boot 1, then matched on boot 2+. PR #330 stopped that pass from racing with `digits-first-boot.service` by adding ordering, but the rewrite itself was still pure boot-time work that landed the same content the rsync layer below already put in place.

## Fix

Two pieces in `tools/build-image.sh`:

1. **Layer `embed/rootfs/` over the rootfs partition** at build time. embed/rootfs/ is generated from rootfs-overlay/ by `make embed`, so files shared between them are no-ops. The only meaningful addition is the cross-compiled `digits-setup` binary, which used to reach the rootfs only via the runtime extractor. Now it lands at flash time.
2. **Pre-write `/data/digits/asset-version`** with the same `version:hash` marker digitsd would compute on first boot. New helper `tools/compute-asset-marker.py` mirrors the Go `Extractor.hashEmbeddedFS` algorithm: sha256 over (sorted path || NUL || data || NUL) for every file in `embed/rootfs/` and `embed/data/`. Cross-checked against a Go reference impl reading the same embed dir; same hash byte-for-byte.

The recovery skeleton tar captures the pre-written marker (it is created from /data after we write asset-version), so factory resets also restore a state where digitsd skips extraction.

## Why this is `chore(image)` not `fix(image)`

Nothing was broken after PR #330 landed: digitsd's runtime extraction worked, just did unnecessary rewrites on every fresh flash. This removes a redundant boot-time step and shrinks the surface area for future races against any other startup work that touches the rootfs.

## Test plan

- [ ] `make image-v2-flash` (dev image)
- [ ] On boot 1, `journalctl -u digitsd | grep 'assets:'` shows `marker matches (version=dev hash=...), skipping extraction` (not `extracting assets for version dev`)
- [ ] `journalctl -b 0 | grep -E 'remount,(rw|ro)'` returns nothing from digitsd (only first-boot's pair, if first-boot still has hostapd sed)
- [ ] `/data/digits/asset-version` exists at flash time, owned by `digits:digits`, mode 644
- [ ] `/usr/local/bin/digits-setup` exists on the fresh rootfs (verifies the embed rsync worked)
- [ ] OTA regression check: bump digitsd `version.Version` and change one byte in any embedded asset, deploy via the updater, confirm the runtime extractor logs `extracting assets for version <new>` and writes the new marker
- [ ] Negative test: `sudo rm /data/digits/asset-version`, reboot, confirm runtime extractor reruns and rewrites the marker without breakage

### DIFF
--- a/pi/image/entrypoint.sh
+++ b/pi/image/entrypoint.sh
@@ -98,12 +98,9 @@ GOOS=linux GOARCH=arm64 go build \
     -o /digits/tools/build/digitsd \
     ./cmd/digitsd/
 
-# Cross-compile digits-setup (pure Go, no CGO needed)
-info "Cross-compiling digits-setup for aarch64..."
-cd /digits/pi/digits-setup
-CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build \
-    -o /digits/tools/build/digits-setup \
-    ./cmd/digits-setup/
+# digits-setup is cross-compiled by `make embed` on the host (see
+# pi/digitsd/Makefile) and lands on the rootfs via build-image.sh's
+# embed/rootfs/ rsync. Nothing to do here.
 
 # Cross-compile digits-recovery (pure Go, no CGO needed). build-image.sh reads
 # this from pi/digits-recovery/bin/ to match the local Makefile output path.
--- a/tools/build-image.sh
+++ b/tools/build-image.sh
@@ -30,6 +30,8 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
 OVERLAY_DIR="${REPO_DIR}/pi/image/rootfs-overlay"
+EMBED_DIR="${REPO_DIR}/pi/digitsd/internal/assets/embed"
+ASSET_MARKER_TOOL="${SCRIPT_DIR}/compute-asset-marker.py"
 PARTITION_SETUP="${REPO_DIR}/pi/image/partition-setup.sh"
 INIT_DATA="${REPO_DIR}/pi/image/init-data.sh"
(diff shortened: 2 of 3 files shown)


justinlindh:
@claude

justinlindh:
Good catch, fixed in b72e2f24. Added `python3` to the `require_cmd` list at line 383 so a missing host binary now fails with the script's standard `die "Required command not found: python3"` instead of bash's bare `command not found`.

On the digits-setup-installed-twice observation: skipping. The redundancy is harmless (both code paths deliver functionally equivalent binaries), and consolidating would mean coordinating changes across `build-image.sh` and `entrypoint.sh` for at most a couple seconds of saved build time. YAGNI.

justinlindh:
Done in f6e009cd. Removed the parallel `digits-setup` cross-compile from `pi/image/entrypoint.sh` plus its preflight check and `install` line in `build-image.sh`. `make embed` is now the only path that produces the binary, and the embed rsync at line 633 is the only path that lands it on the rootfs, so the on-disk `/usr/local/bin/digits-setup` byte-for-byte matches what's hashed into the asset-version marker.

justinlindh:
@claude - does that address your concerns? Re-review, por favor.

PR #338, opened 2026-04-27
fix(digitsd): poll PING after reflash so first POST cleanly passes

justinlindh:
## What

Replace the single `Ping()` after reflash with a 10 s polling retry, so a successful first-boot Pico flash no longer leaves a misleading `POST: Continuing without Pico. Phone will not function.` line in the journal.

## Why

Observed on the working `192.168.2.28` carrier swap (PR #335 merge plus a fresh `make image-v2-flash`). The factory-fresh-flash flow added in #334 worked end-to-end, but the journal said:

```
reflash: flash script succeeded reason=post-fail
TX cmd=PING
reflash: PING failed after flash error="serial: timeout waiting for response to PING"
POST: Continuing without Pico. Phone will not function.
```

`flash-pico.sh` already sleeps twice for 2 s during its own flow, and `reflashPico` adds another 2 s before opening the serial port. That's usually enough, but a virgin Pico cold-booting freshly written firmware sometimes isn't ready in time for a single `Ping()` whose internal timeout is 2 s. The next systemd cycle catches up and POST passes, so it's self-healing, but the message is a lie during the gap.

Polling with 500 ms backoff up to 10 s gives ~4 attempts. A healthy Pico is caught immediately; a slow one within a couple seconds. Same code path runs for both reflash reasons (`post-fail` and `version-mismatch`).

## Test plan

- [x] `go test ./...` in `pi/digitsd` (3 new `TestPollPing_*` cases: immediate success, success on retry, deadline expiry)
- [ ] (manual, next factory-fresh flash) confirm journal shows `reflash: PING PASS` instead of `Phone will not function` after the first auto-flash

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -1186,12 +1186,37 @@ func reflashPico(sp *phone.SerialPort, serialDev string, serialLogger *slog.Logg
 	if err != nil {
 		log.Fatalf("reflash: serial re-open failed: %v", err)
 	}
-	if err := newSp.Ping(); err != nil {
-		slog.Warn("reflash: PING failed after flash", "error", err, "reason", reason)
+	// A virgin Pico needs to cold-boot the freshly written firmware before
+	// it can answer PING; flash-pico.sh's own sleeps don't always cover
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #340, opened 2026-04-27
chore(server): drop SWD chip from phone detail

justinlindh:
## Summary
- All shipped devices are flash-capable, so the SWD indicator chip is now noise. Drop it from the firmware row on the phone detail page.
- Drop the now-dead `{{if .DeviceInfo.FlashCapable}}` branches around the firmware update disclosure and its "requires SWD wiring" fallback copy.
- Drop the matching ` · SWD` suffix from the firmware row on the answering-machine theme so all three themes (intercom, dialup, answering-machine) are clean.

## Test plan
- [x] `go build ./cmd/signald/` succeeds
- [x] `go vet ./internal/web/...` clean
- [x] `go test ./internal/web/...` passes
- [ ] Eyeball `/phones/<n>` under each theme (intercom, dialup, answering-machine) to confirm the firmware row no longer shows SWD and the firmware update disclosure still renders

---
_Generated by [Claude Code](https://claude.ai/code/session_01JUYmDPCtkzi2B6LB4ru2CX)_


justinlindh:
@claude

PR #342, opened 2026-04-28
fix(digitsd): keep easter eggs from eating service-code digits

justinlindh:
## Summary

- \`*#00000#\` triggered the Rick Roll easter egg instead of factory reset because the easter-egg trigger \`0000\` is a substring of the code.
- The dispatch loop fed each key to the easter-egg detector first and short-circuited the service-code handler on a match, so the 4th zero of \`*#00000#\` fired Rick Roll and the service-code buffer never saw the full sequence.
- Add \`ServiceCodeHandler.InCode()\` (true when the buffer starts with \`*\`) and gate the easter-egg check on \`!InCode()\` in the digitsd dispatch loop. All service codes begin with \`*\`, so once the user types it they're committed to a code path until completion or hang-up.

## Trace before fix

Keys: \`*\`, \`#\`, 0, 0, 0, 0, 0, \`#\`

| Key | easter buf | svc buf | Result |
|---|---|---|---|
| \`*\` | \`*\` | \`*\` | no match |
| \`#\` | \`*#\` | \`*#\` | no match |
| \`0\` | \`*#0\` | \`*#0\` | no match |
| \`0\` | \`*#00\` | \`*#00\` | no match |
| \`0\` | \`*#000\` | \`*#000\` | no match |
| \`0\` | matches \`0000\` → fires Rick Roll, buffer cleared | \`*#000\` (svc skipped) | Rick Roll plays |
| \`0\` | \`0\` | \`*#0000\` | no match |
| \`#\` | \`0#\` | \`*#0000#\` (7 chars) | factory reset never fires |

## Test plan

- [x] \`go test ./internal/phone/\` passes (new \`TestServiceCodeInCode\` and \`TestFactoryResetVsRickRollEasterEgg\` regression test that mirrors the dispatch logic)
- [x] \`go build ./...\`, \`go vet ./...\`, \`golangci-lint run ./...\` clean from \`pi/digitsd\`
- [ ] Hardware verification on a phone: dial \`*#00000#\` → factory reset triggers, Rick Roll does not play
- [ ] Hardware verification: dial \`0000\` from idle → Rick Roll still plays

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -2076,8 +2076,12 @@ func main() {
 						})
 					}
 				}
-				// Check easter eggs, then service codes
-				if !easterEggs.AddKey(key) {
+				// Mid-service-code: route the key only to svcCodes so easter
+				// eggs (e.g., "0000" Rick Roll) cannot eat digits belonging
+				// to a code like "*#00000#" (factory reset). Otherwise: try
+				// easter eggs first, then fall through to service codes.
+				inCode := svcCodes.InCode()
+				if inCode || !easterEggs.AddKey(key) {
 					switch svcCodes.AddKey(key) {
 					case phone.ServiceCodeTerminal:
 						ctrl.Reset()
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

justinlindh:
Addressed the false-suppression note in #343: tightened `InCode()` to require the full `*#` prefix instead of just `*`, so a lone `*` followed by digits stays in normal-dial mode and easter eggs (`*0000` → Rick Roll) still fire as expected. Regression test included.

PR #343, opened 2026-04-28
fix(digitsd): tighten service-code prefix to "*#" so lone '*' doesn't gate eggs

justinlindh:
Follow-up to #342, addressing the [false-suppression note](https://github.com/justinlindh/digits/pull/342#issuecomment-4331822561) from the review on that PR.

## Summary

- After #342, \`InCode()\` returned true the moment the buffer started with \`*\`, suppressing the easter-egg detector for everything that followed. A lone \`*\` from idle plus four zeros (\`*0000\`) is normal-dial input, not a service code, and used to fire Rick Roll. With \#342 it stopped firing.
- Every real service code begins with the two-character prefix \`*#\`. Tightening \`InCode()\` to require both characters keeps the original bug fix (\`*#00000#\` → factory reset, no Rick Roll) while restoring pre-#342 behavior for lone-\`*\` sequences.
- Added \`TestEasterEggFiresAfterLoneStar\` as a regression for that case; tightened \`TestServiceCodeInCode\` to cover the new boundary.

## Test plan

- [x] \`go test ./internal/phone/\` passes (full suite, plus new lone-\`*\` regression)
- [x] \`go build ./...\`, \`go vet ./...\`, \`golangci-lint run ./...\` clean from \`pi/digitsd\`
- [ ] Hardware verification on a phone: dial \`*#00000#\` → factory reset still triggers, no Rick Roll
- [ ] Hardware verification: dial \`*0000\` → Rick Roll plays
- [ ] Hardware verification: dial \`0000\` from idle → Rick Roll plays

### DIFF
--- a/pi/digitsd/internal/phone/servicecodes.go
+++ b/pi/digitsd/internal/phone/servicecodes.go
@@ -113,12 +113,14 @@ func (h *ServiceCodeHandler) Reset() {
 	h.buffer = ""
 }
 
-// InCode reports whether the user is mid-entry of a service code. All service
-// codes begin with '*', so a buffer starting with '*' means a code is in
-// progress. Callers use this to suppress competing key consumers (e.g., the
-// easter-egg detector) that would otherwise eat keys belonging to the code.
+// InCode reports whether the user is mid-entry of a service code. Every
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #349, opened 2026-04-28
chore(server): drop dead FlashCapable plumbing post-#340

justinlindh:
## Summary

Holistic-review finding from the 2026-04-28 across-PR pass.

PR #340 (`chore(server): drop SWD chip from phone detail`) removed both UI consumers of `DeviceInfo.FlashCapable`: the SWD chip on the phone-detail page and the `{{if .DeviceInfo.FlashCapable}}` gate around the "Update firmware" disclosure. The PR description was explicit: "All shipped devices are flash-capable, so the SWD indicator chip is now noise."

The field was nonetheless left plumbed through:
- `server/internal/signaling/protocol.go` (Message struct)
- `server/internal/signaling/hub.go` (Conn struct, DeviceInfoSnapshot, UpdateDeviceInfo signature)
- `server/internal/signaling/relay.go` (dispatch)
- `server/README.md` message-types table

After #340 there is no server-side consumer that reads the value.

## What this PR does

Drops `FlashCapable` from server-side code and docs. The two test/handler call sites for `UpdateDeviceInfo` are updated to drop the trailing `false` argument.

digitsd keeps emitting the `flash_capable` JSON field; the server's JSON decoder silently ignores unknown fields on unmarshal, so:
- Old digitsd to new server: server ignores `flash_capable` (fine).
- New digitsd to old server: digitsd still sends it (fine).
- A subsequent digitsd-side cleanup can drop the wire field once cross-version concerns clear.

## Motivated by (last 24h merges)

- #340 chore(server): drop SWD chip from phone detail (the explicit "noise" call)

## Test plan

- [x] `go build ./...` clean (server module, Go 1.26.1 via auto-toolchain)
- [x] `go test ./internal/signaling/ ./internal/web/` pass
- [x] `go vet ./...` clean
- [ ] CI-side `golangci-lint` (host binary is built against Go 1.25 and refuses 1.26.1 modules)

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -228,7 +228,7 @@ After SDP/ICE exchange, audio flows peer-to-peer via WebRTC DTLS-SRTP.
 | Type              | Direction      | Key Fields                            | Purpose                              |
 |-------------------|----------------|---------------------------------------|--------------------------------------|
 | `register`        | Phone -> Server | `number`                             | Register phone number on connect     |
-| `device_info`     | Phone -> Server | version fields, `flash_capable`      | Report device version info           |
+| `device_info`     | Phone -> Server | version fields                       | Report device version info           |
 | `call`            | Phone -> Server | `to`                                 | Initiate call to another number      |
 | `ring`            | Server -> Phone | `from`                               | Notify callee of incoming call       |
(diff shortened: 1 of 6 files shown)


justinlindh:
@claude

PR #353, opened 2026-04-28
fix(server): thread ctx through admin and updates HTTP clients

justinlindh:
`StatsClient.Fetch` and `GitHubReleases.fetchSHA256` accepted a `context.Context` but called `http.NewRequest`, which silently dropped it. Caller cancellation and deadlines had no effect on the in-flight HTTP request.

Switching both to `http.NewRequestWithContext` makes the ctx parameter actually do what its presence in the signature implies.


justinlindh:
@claude

PR #354, opened 2026-04-28
refactor(server): share Call row scan between Recent and RecentForPhones

justinlindh:
`Recent` and `RecentForPhones` had byte-identical 11-column SELECT lists, identical row-scan args, and identical `sql.NullString` to `*string` translation for `force_ended_by`. A schema change required updating both in lockstep.

Extracted a `callColumns` const for the SELECT list and a `scanCallRows(*sql.Rows)` helper for the loop. `recentCallsForHistory` (intentionally narrower 10-column SELECT, no `force_ended_by`) is left alone.

### DIFF
--- a/server/internal/calls/tracker.go
+++ b/server/internal/calls/tracker.go
@@ -371,17 +371,12 @@ func (t *Tracker) Active() []activeCall {
 	return calls
 }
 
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

PR #355, opened 2026-04-28
refactor(server): share Line row scan across List variants

justinlindh:
`List`, `ListByHousehold`, and `ListByHouseholds` each repeated the same 7-column SELECT, the same `rows.Scan` arg list, and the same `scanSettings` post-processing. A column add or rename meant editing three near-identical loops in lockstep.

Extracted a `lineColumns` const and a `scanLineRow` helper. Single-row callers (`Add`, `GetByID`, `GetByNumber`) keep their existing inline scans because they each have distinct error wrapping and `sql.ErrNoRows` handling that doesn't fit the shared helper cleanly.

### DIFF
--- a/server/internal/line/store.go
+++ b/server/internal/line/store.go
@@ -33,6 +33,24 @@ func scanSettings(raw []byte) (Settings, error) {
 	return DefaultSettings().Merge(patch).Normalize(), nil
 }
 
+// lineColumns is the SELECT list shared by queries that scan rows into a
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

justinlindh:
Done. Trimmed the keep-in-sync comment; the const now sits bare above `scanLineRow` and the proximity carries the coupling. Fixed in 07936004.

PR #356, opened 2026-04-28
refactor(server): collapse RevokeLink into a single UPDATE...RETURNING

justinlindh:
`RevokeLink` ran a SELECT into `householdAID`/`householdBID`, never used those values, then ran an UPDATE inside a transaction that wrapped only those two statements. The doc comment claimed the function cascade-deletes contacts between linked households, but that has never happened in the implementation: contacts cascade only on `phones` delete, not on link revoke.

Replaced with a single `UPDATE ... WHERE ... AND status != 'revoked' RETURNING id`. The "link not found or already revoked" error path is preserved via `sql.ErrNoRows` from the RETURNING. Test (`TestRevokeLink_ChangesStatus`) still passes.

### DIFF
--- a/server/internal/household/link.go
+++ b/server/internal/household/link.go
@@ -175,40 +175,23 @@ func (s *LinkStore) AreLinked(ctx context.Context, householdAID, householdBID st
 	return count > 0, nil
 }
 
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude please review this PR

PR #358, opened 2026-04-28
chore(ci): dial up the release-groomer voice

justinlindh:
## What

Refines the release-voice system prompt that `.github/workflows/release-groom.yml` appends when it grooms each release body. Same hard rules (no em-dashes, 120-word cap, sentinel-first output, no per-item headings); the voice section is rewritten and the three examples updated.

## Why

The previous voice was warm-dry-cheeky and produced perfectly fine notes. We wanted more bite: openly mock-celebratory, "oh wow, the nerds shipped some code, marvel at their works." The new prompt explicitly calibrates per change type (mock-fanfare for features, gentle shaming for bug fixes, minimum effort for chores) and adds tonal guardrails so the joke never buries the changelog content or punches at users.

## Test plan

Dry-ran the new prompt against four real releases using the same model (`claude-sonnet-4-6`) and the same `--append-system-prompt` mechanism the workflow uses. Outputs below; existing groomed bodies shown for comparison where they exist.

### fw/v1.8.0 (firmware, big release)

**Before:**
> Sensitive service codes now ask for a star-press before they fire. Two bugs went down with that fix: easter eggs were quietly eating service-code digits, and a bare star was enough to trip the confirmation gate. The dial tone had developed a habit of going silent after hangup and after volume or audio-test codes. It speaks again. *#SETUP# was also fixed to actually clear the wifi-configured flag, which it was not. On the web side: the call log paginates across all three themes, the answering-machine dashboard updates its counters and clock live, and you can rename a line right from the detail page.

**After:**
> Hold onto your handset: sensitive service codes now require a deliberate star-press confirmation before they'll fire. While fixing that, two related bugs got swept up: easter eggs were eating service-code digits, and a lone star was enough to trip the gate on its own. The dial tone works again in the two situations where it had stopped working: after hangups, and after volume or audio-test codes. *#SETUP# now actually clears the wifi-configured flag, which it had been quietly declining to do. On the web side, the call log paginates across all three themes, the answering-machine dashboard updates its counters and clock live, and you can rename a line right from the detail page.

### pi/v1.16.0 (pi software)

**Before:**
> Sensitive service codes now require a press-star confirmation before they fire, so a pocket dial can no longer reboot your phone into setup mode. When you do trigger AP mode intentionally, the phone says "rebooting into AP mode" out loud before it goes dark, which is a small but welcome heads-up. Also fixed: *#SETUP# was not actually clearing the wifi-configured flag, meaning re-setup would silently fail. Now it does what it says on the tin.

**After:**
> Hold onto your handset: sensitive service codes now demand a star-key confirmation before doing anything drastic, meaning a pocket dial can no longer reboot your phone into setup mode. When you do trigger AP mode intentionally, the phone announces "rebooting into AP mode" before it goes dark. The engineers have apparently discovered that phones should tell you what they're doing. Also: the server now clears stale pairing when a phone initiates a repair, so re-pairing no longer silently fails. Invite codes are generated uppercase now, matching the part of the code that was already upcasing them before comparing. Airtight engineering, as always.

### server/v1.44.0 (small release, never groomed: raw semantic-release body)

**Before:** raw semantic-release markdown (\`### Features\` + bullet list).

**After:**
> Breaking news from the nerd factory: the call log is now paginated across all three themes. If your call history was long enough to scroll into the void, it now loads in proper pages instead. You're welcome.

### server/v1.45.0 (also raw, never groomed)

**Before:** raw semantic-release markdown.

**After:**
> Breaking news from the nerd factory: the keypad now asks you to confirm before dialing sensitive service codes, so fat-fingering your way into something unintended is no longer quite so easy. On the pairing front, phones can now tell the server when something has gone sideways and needs a fresh start, which means stale pairing state gets cleaned up automatically instead of haunting you until someone with SSH access fixed it manually.

### DIFF
--- a/.github/release-voice.md
+++ b/.github/release-voice.md
@@ -6,12 +6,34 @@ the raw semantic-release body on the GitHub release.
 
 # Voice
 
-Warm, dry, a little cheeky. Never hype. Never condescending. Write like
-the project's sidebar footer reads, not like a marketing blog.
-Affectionate toward the people building the hardware. If a fix is
-clever, say so plainly ("a nerd in the nerd factory really knocked it
-out of the park"). If a change is boring, either leave it out or be
-boring about it on purpose.
+Mock-celebratory, openly sarcastic, lovingly insulting toward the
+engineers (who are us). The whole vibe is "oh wow, the nerds shipped
+some code, marvel at their works." You are the project gently making
+fun of itself, in public, in front of users. Affectionate, never
+bitter. The target is always the engineers, never the people using
+the phones.
+
+Calibrate per change:
+- Bug fixes are an excuse to gently shame the team for letting the bug
+  exist in the first place. ("Turns out the dial tone was supposed to,
+  you know, work after a hangup. Wild concept. Fixed.")
+- Real new features get mock-fanfare and then a plain-English
+  description of what the feature actually does. ("Hold onto your
+  handset:", "Breaking news from the nerd factory:", etc.)
+- Boring chores get the smallest possible amount of effort. ("The
+  nerds compiled some code. None of it concerns you.")
+
+Hard tonal rules:
+- Information first, jokes second. The user must finish the notes
+  knowing what changed. If a joke is in the way of the information,
+  cut the joke.
+- One mock-fanfare moment per release, max. Don't repeat the bit. If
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude please review this PR

PR #361, opened 2026-04-28
chore(ci): direct-address voice + Silly Goose Award for the release groomer

justinlindh:
## What

Two changes to the release groomer:

1. **Voice file**: replaces the third-person "engineers / nerd factory" framing with explicit second-person address to Justin, the project's sole contributor.
2. **Silly Goose Award**: when Justin ships a bug that is plainly his own fault (forgot to clear a flag, generated lowercase invite codes while the handler does ToUpper, etc.), the groomer can now award him the Silly Goose Award and reference how many days it has been since his last win, with sarcasm calibrated to the gap size.

The workflow scans prior release bodies for the literal phrase "Silly Goose Award", picks the most recent one strictly before this release's publish date, and passes the day gap to the model as a "Silly Goose context:" line. The model decides whether the release deserves the award (one per release max, only for clear Justin-fault bugs, never for routine debugging or upstream regressions). If it does not award, it ignores the gap entirely.

## Why

The previous voice landed but kept the joke at arm's length: an anonymous "team" had quietly shipped a bug, "the nerds" had compiled some code. Justin asked for it to be more direct and more playful. The Goose Award gives the groomer a recurring bit it can deploy with rising sarcasm based on real state in the repo, instead of just stringing together one-liners.

## Test plan

Dry-ran the new prompt against five representative releases with hand-crafted Goose contexts, using the same model (`claude-sonnet-4-6`) and `--append-system-prompt` mechanism the workflow uses.

The 45.0 case is the load-bearing one: the model honored the rule that "no Justin-fault bug = no award, regardless of gap." Clean silent skip.

### DIFF
--- a/.github/release-voice.md
+++ b/.github/release-voice.md
@@ -6,34 +6,49 @@ the raw semantic-release body on the GitHub release.
 
 # Voice
 
-Mock-celebratory, openly sarcastic, lovingly insulting toward the
-engineers (who are us). The whole vibe is "oh wow, the nerds shipped
-some code, marvel at their works." You are the project gently making
-fun of itself, in public, in front of users. Affectionate, never
-bitter. The target is always the engineers, never the people using
-the phones.
+You are talking directly to Justin, the sole person who has ever
+contributed to this project. Address him by name, in the second
+person ("you", "your"), and give him a hard time about whatever just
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #367, opened 2026-04-29
feat(pi): ship V2 mixer state via OTA, render at digitsd startup

justinlindh:
## What

Embed both per-PCB-variant ALSA mixer state files into digitsd and render the matching one to `/data/digits_mixer.state` at startup. The per-codec canonical state now ships through the existing OTA channel (digitsd binary), not just through fresh SD-card flashes.

## Why

PR #366 raised the V2 baseline volume by editing `digits_mixer_v2.state`, but the OTA updater only ships two artifacts: the digitsd binary and the Pico firmware ELF. The mixer state file is laid down by `tools/build-image.sh` on first flash and never touched by OTA again. So existing V2 phones (Beige V2, etc.) would keep the old quieter HP DAC value (-19 dB) indefinitely after #366 merged. They would silently regress relative to fresh-flashed phones.

The state file change needs to ride OTA. Since other per-PCB-variant assets in this codebase (SWD configs at `main.go:1535`) already use an embed + render pattern, this PR follows that precedent exactly.

## Design

- **Source of truth** moves to `pi/mixer-state/v{1,2}.state`, paralleling `pi/tones/`. The Makefile's `embed:` target copies these into `internal/assets/embed/mixer/` at build time, alongside the rootfs and tones copies that were already there.
- **Embed placement.** `embed/mixer/` deliberately sits outside the prefixes `assets.Extractor` walks (it only walks `rootfs/` and `data/`), so the per-variant files stay in the binary and never auto-extract to disk. They're consumed only by the render step via `assets.SubFS()`.
- **Render step** in `main.go`, right after the SWD render: pick `mixer/v{1,2}.state` from `audio.CodecPCBVariant()`, hash-compare against `/data/digits_mixer.state`, write via the existing sudo-aware `extractor.RootfsWriteFile` when content differs. Idempotent skip when already current.
- **Live apply is intentionally deferred** to next boot. Running `alsactl restore` from digitsd would need a sudoers rule for the digits user, and `etc/sudoers.d/digits-updater` is in the Makefile's `OVERLAY_EXCLUDE` list -- the OTA channel can't update sudoers. `digits-mixer.service` already runs `alsactl restore` at boot before digitsd starts, so it picks up the new file on the next reboot.
- **Drop `alsactl store` from `SetVolume`** in `servicecodes.go`. The embedded copy is now the canonical source for codec gain stages and routing, and `/data/digits/volume` already holds the user-facing level. Snapshotting live mixer state on every volume change was wasted SD writes that the next-boot render would overwrite anyway.

## Trade-off

Mixer state changes shipped via OTA take effect on the **next reboot**, not immediately. This is a strict improvement over the previous behavior (changes required SD reflash). Live volume adjustments via `*#*N` service codes still apply instantly via `amixer cset` -- the audio group permission for that has not changed.

## Test plan

- [x] `go test ./...` from `pi/digitsd/` green
- [x] Deployed to Beige V2 (line 17, hw `630dfae1...`) which had the OLD state file (HP DAC=80) on disk: render wrote new file (HP DAC=100), `systemctl restart digits-mixer.service` applied to live mixer (HP DAC=100). Steady-state subsequent run logs `mixer render: already current card=digitscodec` with no write churn.
- [x] Verified `embed/mixer/v{1,2}.state` are present in the built binary via `TestEmbedHasMixerStateVariants` -- the test skips when the embed dir is empty (`make embed` not run) and asserts presence + valid `state.` header otherwise.


justinlindh:
@claude

PR #369, opened 2026-04-29
fix(server): collapse AM mobile chrome behind a single MENU drawer

justinlindh:
## Summary

The AM theme's mobile topbar was wrapping into two awkward rows: brand + MENU on the first, and the household label + Sign out chicklet stacked underneath in their own band. MENU floated alone next to the small two-line wordmark and read as a disconnected fourth element rather than part of a coherent control strip.

This wraps the tabs nav and the household / DND / Sign out cluster in a single `.am-shell__drawer` that the MENU button toggles. On desktop the drawer is just a flex container that absorbs the remaining width with the same internal alignment as before (no visual change). On mobile the entire drawer collapses behind MENU, so the topbar reads as brand left, MENU right, and tapping MENU drops a panel below containing the tabs, the household label, DND if active, and Sign out.

A11y: `aria-controls` and `aria-expanded` on the toggle are pointed at the new drawer id and verified to flip correctly on click.

### Mobile, menu closed
Mobile closed

### Mobile, menu open
Mobile open

### Desktop (unchanged)
Desktop

## Test plan

- [x] `make build`
- [x] `make test`
- [x] Mobile (390x844): topbar is brand + MENU only; MENU expands to a panel below with tabs + household + Sign out
- [x] Mobile: `aria-expanded` flips false to true on click; `aria-controls` matches the drawer id
- [x] Desktop (1440x900): chrome unchanged (tabs in middle, household + Sign out right-aligned)

### DIFF
--- a/server/internal/web/static/answering-machine.css
+++ b/server/internal/web/static/answering-machine.css
@@ -550,6 +550,10 @@ html, body { margin: 0; padding: 0; font-family: var(--font-body); color: var(--
   flex-wrap: wrap;
 }
 .am-shell__brand-slot { display: inline-flex; align-items: center; gap: 12px; padding-right: 8px; }
+/* Drawer wraps tabs + identity (household, DND, signout). On desktop it just
+   absorbs the remaining width; on mobile the whole thing collapses behind
+   the MENU toggle so the topbar stays brand+MENU. */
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

PR #370, opened 2026-04-29
fix(server): cache-bust static assets via commit-suffixed query params

justinlindh:
## Summary

Cloudflare caches `/static/*.css` and the other static assets for 4 hours (`max-age=14400`). After PR #368 and #369 deployed, the new HTML referenced new class names (`am-shell__drawer`, `am-update__row`, etc.) but Cloudflare kept serving the pre-merge CSS to anyone who hit the edge cache, so:

- The AM mobile MENU button toggled `aria-expanded` and added `.is-open` to a div the cached CSS didn't know how to style: nothing happened visually.
- The Software update plate on phone-detail mobile collapsed without the column reorientation, so the INSTALL chicklet, "Up to date" lamp, and version well piled on top of each other.

Both issues were CSS staleness on a stable URL, not a code bug.

## Fix

- New `{{static "path"}}` template helper returns `/static/path?v=<commit-sha>`. Cloudflare keys cache entries on the full URL including query, so each release writes URLs the CDN has never seen, gets a clean cache miss, fetches fresh content. The prior URL's cache entry becomes a tombstone that ages out harmlessly.
- All `<link>` and `<script>` refs in the four layouts (`layout-v2`, `layout-dialup`, `layout-answering-machine`, `connecting`) plus `welcome.html` updated to use the helper. Body-section `` and `<audio>` refs left bare (stable assets, not part of the breakage).
- `staticFileServer` now sets `Cache-Control: public, max-age=31536000, immutable` when the request carries a query string, so the fingerprinted URLs can cache aggressively at the edge while unversioned fallbacks (dev mode, where `version.Commit` is `unknown`) revalidate per request.
- `baseTemplateFuncs` extracted to a package-level helper exposed via `TemplateFuncs()` so the two test files and `cmd/signald/main.go` (which parse `layout-v2.html` outside `NewHandler`) share the same funcMap. Without that, the binary panicked on startup with `function "static" not defined`.

## What about the currently-cached entries?

The OLD `/static/answering-machine.css` URL is still cached at Cloudflare for the rest of its 4-hour TTL. After this PR deploys:
- New page loads will reference `/static/answering-machine.css?v=<NEW-SHA>`, a cache miss, fresh content. ✅
- Anyone holding stale HTML in an open tab will continue requesting the bare URL and getting the stale CSS until they refresh.

Optional one-time Cloudflare cache purge of `/static/*` would clear the tombstones immediately. Not strictly required.

## Test plan

- [x] `make build`
- [x] `make test`
- [x] Dev server: rendered `<head>` for `/login`, `/welcome`, `/links` (AM theme) all show `?v=<commit>` on every CSS/JS/font/favicon ref
- [x] Versioned asset request returns `Cache-Control: public, max-age=31536000, immutable`
- [x] Bare asset request returns no `Cache-Control` (default short-lived)

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -115,7 +115,7 @@ func run(ctx context.Context) error {
 		slog.Info("Google OAuth enabled")
 	}
 
-	loginTmpl, err := template.ParseFS(web.TemplateFS(), "templates/layout-v2.html", "templates/_partials.html", "templates/login.html")
+	loginTmpl, err := template.New("").Funcs(web.TemplateFuncs()).ParseFS(web.TemplateFS(), "templates/layout-v2.html", "templates/_partials.html", "templates/login.html")
 	if err != nil {
 		return fmt.Errorf("parse login template: %w", err)
 	}
--- a/server/internal/web/e2e_auth_test.go
+++ b/server/internal/web/e2e_auth_test.go
@@ -42,7 +42,7 @@ func setupAuthTestServer(t *testing.T) *httptest.Server {
 	emailSender := email.NewNoopSender()
 
 	// Parse login template from the embedded FS
-	loginTmpl, err := template.New("").ParseFS(TemplateFS(), "templates/layout-v2.html", "templates/_partials.html", "templates/login.html")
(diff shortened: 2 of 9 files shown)


justinlindh:
@claude

PR #261, opened 2026-04-23
feat(server): 3-way link-health host-gated kick (phase 4 of 4)

justinlindh:
## What

Phase 4 of 4: host-gated per-member kick on the 3-way conference observation deck.

- Schema v22 adds the `conference_kicks` audit table, cascade-tied to `conferences`, append-only.
- `Tracker.RecordKick` writes the audit row.
- `requireConferenceHostOwnership` 404s unless the authenticated user's household directly owns the conference host phone.
- `POST /api/conference/{uuid}/kick` wires auth, audit, kick, and SSE notify. v1 conference semantics collapse the whole conference on any drop, so remaining members flow through the existing `dropMemberFromConference` and the surviving pair's continuation call is created by the existing persistence layer.
- `Relay.KickMember` sends `TypeConferenceEnd` to the kicked phone so it tears down cleanly, then reuses `dropMemberFromConference` for the remaining-pair notifications.
- `HealthStore.NotifyDisconnectedConference` fans out a `DisconnectKind` event to SSE observers with the actor's display label; the deck panel flips to "Conference ended by <name>" before the evict cascade closes the channel. Added via a shared `notifyDisconnectedSession` helper extracted from the existing 2-party `NotifyDisconnected`.
- Deck template renders `Remove` buttons on non-host member cards when the viewer's household owns the host phone; click opens a per-member confirm dialog and POSTs to the kick endpoint.

No new signaling message type, no Pi or firmware changes.

## Auth

Tightens the conference gate from "any member household observes" (Phase 3) to "host household only can kick." Member-household observers still see the deck but do not see the kick UI.

## Test plan

- [x] `go test ./...`
- [x] `make test-integration`:
  - `TestConferenceKicksSchemaV22_Integration`, `TestRecordKick_Integration`, `TestKickMember_SendsConferenceEndToKickedAndRemaining_Integration`.
  - `TestRequireConferenceHostOwnership` (5 cases), `TestConferenceKickEndpoint_HostSucceeds`, `TestConferenceKickEndpoint_MemberForbidden`, `TestConferenceKickEndpoint_KickHostReturns400`, `TestConferenceKickEndpoint_EndedConferenceReturns404`.
  - `TestConferenceSSEStream_KickTriggersDisconnectEvent`.
  - `TestConferenceLiveDetailPage_OwnerRenders` (extended with kick-button assertions), `TestConferenceLiveDetailPage_MemberHouseholdNoKickButtons`.
[...]

PR #382, opened 2026-05-02
fix(digitsd): populate loss/jitter from InboundRTPStreamStats

justinlindh:
## Why

`/calls`, `/call/live/{id}`, and `/conference/live/{uuid}` have shown `-` for **loss** and **jitter** on every call since the link-health observation deck shipped (#227 / #237 / #260 / #261). RTT, route, and bytes have always populated correctly.

Root cause is on the phone side: digitsd's link-health reporter type-asserted against `pionwebrtc.RemoteInboundRTPStreamStats` (the remote's RR view of our outbound). Pion v4.x does not emit that type in our codec configuration, so the assertion failed every tick and `LossPct` / `JitterMs` were always nil. The wire payload, server store, and templates were all correct.

## What

Switch the type assertion to `InboundRTPStreamStats` (pion's local view of the inbound stream). Jitter comes from the field directly (seconds to ms). For loss the inbound type carries lifetime `PacketsReceived` / `PacketsLost` counters, so the Reporter now keeps the previous-cycle counters and reports loss as `lostDelta * 100 / (recvDelta + lostDelta)` over the last sample window.

Followup commit on this branch tightens the loss guard with `recvDelta >= 0` (closes a corner where `PacketsReceived` decreases on track replacement), documents the single-goroutine ownership invariant on `Reporter`, refreshes the doc comments that still referred to the old RR path, and adds three guard-branch tests (counter reset, zero-traffic window, all-loss window).

This means the displayed loss/jitter is now **what each phone is hearing**, not what the remote claims about us. That is the metric users actually care about for "how good is my call quality" and matches the existing copy on the deck (`Bars show packet loss and jitter reported by each endpoint`).

## Verification

- digitsd unit tests pass (11 reporter cases covering nominal, RR-only, ICE-only, counter reset, zero-traffic, all-loss windows).
- golangci-lint clean.
- Deployed to two production phones, placed call 119 between lines 245-1234 and 245-7890. Prod DB:
  ```
  call_id | endpoint | n  | loss_n | jit_n | jit_min   | jit_max
  --------+----------+----+--------+-------+-----------+----------
      119 | 2451234  |  5 |      5 |     5 |  7.067957 | 16.259243
      119 | 2457890  |  5 |      5 |     5 | 4.220025  | 14.963619
  ```
  vs prior call 118 on the unfixed binary, same lines: `5 / 0 / 0`.
- `/call/live/119` renders both endpoints' bars with values:

  call 119 detail with populated loss and jitter

## Test plan

- [ ] CI green (digitsd unit, server unit, server integration).
- [ ] After merge + deploy, place a call, confirm `/calls` row links into a detail view with non-dash loss / jitter for both endpoints.
- [ ] 3-way conference observation deck matrix populates per-edge loss / jitter (covered by deployment, no separate code path).

### DIFF
--- a/pi/digitsd/internal/webrtc/linkhealth.go
+++ b/pi/digitsd/internal/webrtc/linkhealth.go
@@ -146,7 +146,7 @@ func (r *Reporter) sample() (Sample, bool) {
 		}
 	}
 
-	// Drop if neither RR nor ICE pair produced anything useful.
+	// Drop if neither InboundRTP nor ICE pair produced anything useful.
 	if out.LossPct == nil && out.JitterMs == nil && out.RttMs == nil &&
 		out.BytesIn == nil && out.BytesOut == nil && out.ConnType == "" {
 		return Sample{}, false
--- a/pi/digitsd/internal/webrtc/linkhealth.go
+++ b/pi/digitsd/internal/webrtc/linkhealth.go
@@ -40,6 +40,10 @@ type Reporter struct {
 	getter   StatsGetter
 	send     func(Sample) error
 	interval time.Duration
+
+	prevPacketsReceived uint32
+	prevPacketsLost     int32
+	havePrev            bool
 }
 
 const defaultInterval = 2 * time.Second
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

justinlindh:
Acknowledged. Both observations are accurate and intentional:

1. Bench test: kept zero counters because the hot path being benchmarked is stats-report iteration + type assertion, not the per-window arithmetic. The 11 unit tests cover the delta path proper.
2. Semantic shift: confirmed correct. Inbound is the right metric for the deck's framing, and the existing UI copy reads cleanly under that interpretation.

Stale doc string fix from /simplify is in commit a3acc472. Merging.

PR #387, opened 2026-05-02
feat(server): show LAN IP for each phone on /phones (owner-only)

### LINKED ISSUE #381: feat(server): show LAN IP for each phone on /phones (owner-only)
## Why

When deploying or troubleshooting from a laptop on the LAN, we currently have no way to map a paired phone (line number) to its actual LAN IP without grepping signald logs. This makes it cumbersome to ssh into the right device or push a one-off binary build to the right pi.

## What

On \`/phones\` (and the per-phone detail page \`/phones/{number}\`), surface the device's LAN IP next to its line/label.

Constraints:
- **Owner-only.** Only show to a user whose household actually owns the phone. Members of *linked* households should not see it. Treat it like the existing pairing-code/setting fields that are owner-gated.
- **LAN-only.** This is the address the WebSocket connection came from on the local network. Public IPs would expose the phone owner's home WAN address to anyone with link access; that's not what this is for.
- We already have the connection's remote address inside signald (the WS \`r.RemoteAddr\` on the device socket). Plumb that through to wherever we render the phone row.

## Acceptance

- Logged-in owner sees \`192.168.x.y\` on the row for each of their paired phones.
- Linked-household users do not see the IP column.
- IP updates when a phone reconnects from a new address (NAT / DHCP renewal).
- Stale IPs are not shown when a phone is offline; either suppress the field or label it clearly.

Out of band of any current work. Filing now so it doesn't get lost.

justinlindh:
## Summary

Surfaces each paired phone's LAN IP on `/phones` and `/phones/{number}`, owner-scope only, present-tense only. Closes #381.

- New `server/internal/httputil` package with `ClientIP` (XFF first, RemoteAddr fallback, host-only) and `IsPrivateAddr` (RFC1918 + loopback + link-local + ULA). The rate limiter now consumes both helpers instead of carrying its own copy.
- `signaling.Conn.RemoteAddr` is set at WS upgrade time, suppressed to `""` for any non-private address (the LAN-only filter), and surfaced through the existing `Hub.DeviceInfo` snapshot. The snapshot's new field carries `json:"-"` and a doc comment marking it owner-scope HTML only; the `TestDeviceInfoSnapshotJSONOmitsRemoteAddr` test pins that contract.
- All three themes (intercom, dialup, answering-machine) render the IP on both the list and detail pages, gated by `online && DeviceInfo.RemoteAddr != ""`. Layout fix on the mobile card: the IP nests inside `lines__num` so the four-column grid stays four-children.
- Integration tests cover `/phones` rendering (set / empty / offline), `handleWS` capture (private XFF, public XFF, loopback fallback), and a defense-in-depth pin that the dashboard does not render the IP.

## Threat model note

The "linked households don't see this" part of the original issue framing is operational ("not a surface they need"), not privacy. Linked households already learn each other's LAN and public addresses via WebRTC ICE candidate exchange during a call. The `json:"-"` tag and the dashboard pin are defense in depth against accidental future serialization, not a meaningful new privacy boundary. Surfacing peer IPs in call logs / link-health is a separate idea worth its own issue if interesting.

## Test plan

- [x] Unit: `httputil.{ClientIP,IsPrivateAddr}`, `signaling.Hub.DeviceInfo` round-trip, JSON-omit pin.
- [x] Integration: `/phones` rendering (3 cases), `handleWS` capture (3 cases), dashboard pin.
- [x] `make test` and `golangci-lint run ./...` clean.
- [x] `make test-integration` (full suite, ephemeral DB, serial) green.
- [ ] Per-theme `agent-browser` screenshots against prod. **To be added after a worktree deploy.**

## Screenshots

To be added after deploying the branch to the GPU box.

### DIFF
--- a/server/internal/httputil/clientip.go
+++ b/server/internal/httputil/clientip.go
@@ -1,6 +1,6 @@
 // Package httputil provides small request-scoped helpers shared across
-// signald handlers. Today it carries client-IP resolution and a
-// private-address predicate.
+// signald handlers, such as client-IP resolution and a private-address
+// predicate.
 package httputil
 
 import (
--- a/server/internal/httputil/clientip.go
+++ b/server/internal/httputil/clientip.go
@@ -1,0 +1,74 @@
+// Package httputil provides small request-scoped helpers shared across
+// signald handlers. Today it carries client-IP resolution and a
+// private-address predicate.
+package httputil
+
+import (
+	"net"
+	"net/http"
+	"strings"
+)
+
+// ClientIP returns the resolved client IP for an inbound HTTP request.
+//
+// Preference order:
+//  1. The first entry of X-Forwarded-For, trimmed.
+//  2. Host portion of r.RemoteAddr (port stripped).
+//  3. r.RemoteAddr verbatim if SplitHostPort cannot parse it.
+//
+// Assumption: a single trusted reverse proxy (nginx-proxy-manager) is the
+// only writer of X-Forwarded-For. Behind that assumption the leftmost
+// XFF entry is the original client. If the deployment grows additional
+// proxies, this function must be revisited.
+func ClientIP(r *http.Request) string {
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
+		parts := strings.SplitN(xff, ",", 2)
+		return strings.TrimSpace(parts[0])
+	}
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		return r.RemoteAddr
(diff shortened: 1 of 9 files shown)


justinlindh:
@claude please review this PR end-to-end. Flag any blockers, security/threat-model gaps, or test gaps. Skip nits.

justinlindh:
Acknowledged. Threat-model assessment matches the PR body's reasoning, and the data-flow trace confirms the dashboard pin / json:"-" tag close the only realistic accidental-leak paths. Merging.

PR #388, opened 2026-05-02
fix(digitsd): survive signald connect failure so wifi-fallback can run

justinlindh:
## Summary

When a wrong WiFi password is entered during setup, NM fails and digitsd's
signaling dial fails with a DNS error. `log.Fatalf` on that failure killed
digitsd before the wifi-fallback supervisor could start, permanently wedging
the device with no path back to AP/setup mode.

- **Signal client**: `Connect()` now closes the `done` channel on failure so
  the main loop's reconnect case fires immediately.
- **digitsd main**: initial `sig.Connect()` failure is non-fatal. The existing
  reconnect loop (exponential backoff, max 60s) handles retry.
- **Boot ordering**: wifi-fallback supervisor and boot-counter clear are moved
  before `sig.Connect()` so they always execute regardless of network state.
- **digits-ap-check**: `do_station_up` no longer clears the boot counter. That
  early clear defeated the initramfs boot-counter recovery path (counter was
  reset to 0 on every boot before digitsd could fail, so the threshold of 3
  was never reached).

Root-caused from SD card journal analysis of a device stuck in this state:
NM logged `no secrets: No agents were available for this request`, digitsd
crash-looped at restart counter 61, boot counter stuck at 0.

## Test plan

- [x] `go vet ./...` passes
- [x] `golangci-lint run ./...` passes (0 issues)
- [x] All existing signal client tests pass
- [x] New `TestClient_ConnectFailureClosesDone` verifies done channel closes
- [x] All wififallback supervisor tests pass
- [x] Full `go test ./...` passes (all packages)

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -2168,12 +2168,18 @@ func main() {
 		requestICEServers(sig)
 	}
 
-	// Refresh ICE credentials periodically (TURN creds are time-limited)
+	// Refresh ICE credentials periodically (TURN creds are time-limited).
+	// Read cb.sig under cb.mu so we use the current client after reconnects.
 	go func() {
 		ticker := time.NewTicker(30 * time.Minute)
 		defer ticker.Stop()
 		for range ticker.C {
-			requestICEServers(sig)
+			cb.mu.Lock()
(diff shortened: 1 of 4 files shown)


justinlindh:
@claude

PR #391, opened 2026-05-03
fix(server): self-report LAN IP from device, not WS upgrade

justinlindh:
## Summary

Follow-up to #387: the LAN IP feature shipped, but on the prod box no phone showed an IP. The diagnostic logs (deployed temporarily on a debug branch) confirmed why.

Phones reach `app.digits.family` through Cloudflare. CF cannot see behind the home LAN NAT, so the XFF chain that arrives at signald looks like:

```
xff:              "68.224.37.131, 162.158.244.138"   user's WAN, CF edge
cf_connecting_ip: "68.224.37.131"                    same; CF authoritative
raw_remote_addr:  "192.168.1.75:39720"               LAN IP of cloudflared/NPM
```

The user's WAN address is correctly rejected by `IsPrivateAddr`, so every phone got `RemoteAddr=""`. The data the feature wants simply isn't on the wire by the time signald sees the request.

## Fix

Make the device the source of truth. `device_info` (sent immediately after register) now carries `local_addr`. digitsd reads its own routable source IP via `net.Dial("udp", "8.8.8.8:80")` (no packet sent; the kernel just resolves the route) and reports it. The server runs the value through `httputil.IsPrivateAddr` before storing so a compromised client cannot push a public IP into the operator UI.

`handler_ws.go` no longer captures from the request: that path was always going to be wrong under any non-trivial proxy chain.

## Verified on prod

After deploying this branch (server image `1.50.0-7-g44314b4a`, digitsd binary `1.19.1-1-g44314b4a-dirty` to all three phones):

```
device_info  number=2486881  pi_version=1.19.1-1-g44314b4a-dirty  local_addr=192.168.2.228
device_info  number=2456390  pi_version=1.19.1-1-g44314b4a-dirty  local_addr=192.168.2.142
device_info  number=5793721  pi_version=1.19.1-1-g44314b4a-dirty  local_addr=192.168.2.140
```

The three phones still on `1.19.1` (no `local_addr` in their build) report empty, which is the right degraded mode and resolves itself the next time they update.

## Test plan

- [x] `make test` and `make test-integration` green.
- [x] `golangci-lint run ./...` clean on both modules.
- [x] Three new `TestDeviceInfo_*` tests cover the device_info path: private value stored, public value dropped, omitted field stays empty.
- [x] Verified end to end on the live prod stack.

## Notes

- The three phones on the LAN are running this binary out of band (deployed before the PR landed) so I could verify. They will pick up the official build whenever the next pi/digitsd release goes out.
- The `httputil.{ClientIP,IsPrivateAddr}` package stays. `IsPrivateAddr` is what filters the device-reported value, and `ClientIP` is still consumed by `ratelimit`. No code is being orphaned.
- `Conn.RemoteAddr` is now empty until `device_info` arrives, a 1-2 second window after register. Page loads in that window won't show an IP. Acceptable for the "ssh into the phone" use case; the typical access is on a phone that's been up for a while.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"log/slog"
 	"math"
+	"net"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -2817,6 +2818,7 @@ func requestICEServers(sig *sigclient.Client) {
 }
 
 func sendDeviceInfo(sig *sigclient.Client, fwVersion, fwCommit string, flashCapable bool) {
+	localAddr := primaryLocalAddr()
 	if err := sig.Send(&sigclient.Message{
 		Type:            sigclient.TypeDeviceInfo,
 		PiVersion:       version.Version,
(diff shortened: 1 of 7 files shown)


justinlindh:
@claude

PR #396, opened 2026-05-03
refactor(server): drop unused ctx from in-memory Tracker lookups

justinlindh:
`Tracker.Busy`, `CanAddAsHost`, `AllPeersOf`, `PeerOf`, `CallIDForPair`, `CallIDFor`, and `InCall` are pure in-memory map reads. They take `t.mu` and return without ever touching the database. The `context.Context` they accepted went unused in every implementation, and the `mockTracker` and `CallTracker` interface stub matched. The result was an interface signature that lied about whether ctx mattered.

Drop `ctx` from all seven methods, the matching `CallTracker` interface declarations, the relay / conference call sites that were threading ctx through purely to satisfy the signature, the dashboard and ws handler sites, and the unit + integration test mocks.

The persistent / DB-backed tracker methods (`OnCallInitiated`, `OnCallAnswered`, `OnCallEnded`, `ClearByNumber`, `*ConferencePersistent`, `DropMemberPersistent`) keep their `ctx` since they actually use it on the SQL exec/query path.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (pre-existing root-only autodeploy fixture failure unrelated)
- [x] `go vet -tags=integration ./...`
- [x] `golangci-lint run --build-tags=integration ./...`

### DIFF
--- a/server/internal/calls/conference_integration_test.go
+++ b/server/internal/calls/conference_integration_test.go
@@ -30,7 +30,7 @@ func TestTrackerBusyWithConference_Integration(t *testing.T) {
 	if err != nil {
 		t.Fatalf("OnCallInitiated: %v", err)
 	}
-	if !tr.Busy(context.Background(), "5550001") {
+	if !tr.Busy("5550001") {
 		t.Fatalf("expected 5550001 busy from 2-party call")
 	}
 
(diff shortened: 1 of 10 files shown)


justinlindh:
@claude - should we do this, or just keep the `ctx` protos to satisfy "use context everywhere"?

PR #398, opened 2026-05-03
fix(digitsd): keep rootfs ro after asset extraction

justinlindh:
## The bug

After deploying a new `digitsd` binary, five of six phones came up with the rootfs in `rw` state instead of `ro`. The `Remount(false)` call in `Extract()` and `ApplyPiUpdate` logged a WARNING and silently continued, leaving the filesystem permanently writable.

## Root cause (confirmed on hardware)

When a running binary is replaced via `mv` (whether from the deploy-phone skill or OTA), the old binary's inode is unlinked but kept alive by the still-running process's `rw-p` mmap. The Linux kernel rejects `mount -o remount,ro` on ext4 while any process holds a deleted inode open on that filesystem:

```
$ sudo cat /proc/<digitsd-pid>/maps | grep digitsd
... /usr/local/bin/digitsd (deleted)    # rw-p mapping blocks remount,ro
```

Reproduced on-device: `cp + mv` the running binary, then `mount -o remount,ro /` returns "mount point is busy". After `systemctl restart digitsd` (releasing the mmap), the remount succeeds immediately.

The previous troubleshooting note in the deploy-phone skill blamed journald, but the journal files live on `/data` (device `179,4`, not `179,2`), confirmed via `stat /var/log/journal/*/system.journal`. Stopping journald happened to work as a workaround only because it caused digitsd's stdout socket (connected to journald via `StandardOutput=journal`) to break, which terminated the old process and released the mmap.

## What I tried

- Confirmed `sync` alone does not fix the deleted-inode case (the blocker is the open mmap reference, not dirty pages).
- Confirmed small file writes + `remount,ro` work fine when no running binary is being replaced.
- Confirmed the remount succeeds cleanly once the old process exits.

## The fix

**`assets.go`**:
- New `Sync func() error` field on `Extractor`. Called before each remount-ro attempt. Wired to `syscall.Sync` in `main.go`.
- New `RemountReadOnly()` exported method: sync, then `Remount(false)`, up to 4 attempts with 300ms delays. This gives a dying predecessor process time to exit and release its inode.
- `Extract()` now calls `RemountReadOnly()` instead of bare `Remount(false)`. Failure is a hard error (not a WARNING), so the marker is withheld and the next boot retries from scratch.

**`main.go`**: SWD config render path uses `RemountReadOnly()` for the same robustness.

**`updater.go`**: Added a comment explaining why the `remount,ro` at the end of `ApplyPiUpdate` is expected to fail (running process still holds the old inode) and that `Extract()` on the next startup corrects the state.

**`deploy-phone/SKILL.md`**: Fixed the deploy sequence to restart the service BEFORE attempting `remount,ro`, so the old mmap is released first. Corrected the troubleshooting section.

## Test plan

- `go test ./...` from `pi/digitsd/`: all pass.
- `go build ./...` from `pi/digitsd/`: clean.
- `golangci-lint run ./...`: 0 issues.
- New tests: `TestRemountReadOnly_RetriesOnFailure`, `TestRemountReadOnly_SucceedsOnSecondAttempt`, `TestRemountReadOnly_CallsSyncBeforeEachAttempt`, `TestExtract_FailsWhenRemountROFails`.
- On-device: verified rw/ro remount cycle works, confirmed deleted-inode scenario via `/proc/<pid>/maps`, confirmed remount succeeds after service restart.

### DIFF
--- a/.claude/skills/deploy-phone/SKILL.md
+++ b/.claude/skills/deploy-phone/SKILL.md
@@ -49,11 +49,19 @@ For each target phone:
 # 1. Copy binary to /tmp (writable without remount)
 sshpass -p <password> scp <local-binary> <user>@<ip>:/tmp/<binary-name>
 
-# 2. Remount rw, install, remount ro, restart (single SSH session)
-sshpass -p <password> ssh <user>@<ip> 'sudo mount -o remount,rw / && sudo mv /tmp/<binary-name> <install-destination> && sudo chmod 755 <install-destination> && sudo mount -o remount,ro / && sudo systemctl restart <service>'
+# 2. Remount rw, install binary, restart service, then remount ro
+#    Restart BEFORE remount ro -- the running service holds the old binary's
+#    inode via mmap and the kernel rejects remount,ro until that inode is
+#    released. Restarting the service drops the old mmap; remount ro then
+#    succeeds. digitsd's Extract() also re-attempts remount ro on startup.
+sshpass -p <password> ssh <user>@<ip> 'sudo mount -o remount,rw / && sudo mv /tmp/<binary-name> <install-destination> && sudo chmod 755 <install-destination> && sudo systemctl restart <service> && sleep 2 && sudo mount -o remount,ro /'
 ```
 
-If no service restart is needed, omit the `systemctl restart` but **always remount back to ro**.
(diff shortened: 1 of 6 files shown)


justinlindh:
@claude

PR #403, opened 2026-05-03
fix(server): sync navbar DND indicator with phones page toggle

justinlindh:
## What

The phones page DND toggle (PR #395) used htmx to swap only `#dnd-toggle`, leaving the navbar indicator stale until a full page reload. The settings page toggle did a full POST redirect so the navbar updated naturally, but the two surfaces felt disconnected: toggling DND on `/phones` changed state without visual feedback in the chrome.

## Changes

- Extract the navbar DND indicator into a shared `dnd-indicator-content` partial in `_partials.html`, rendered by all three layout templates inside a stable `<span id="dnd-indicator">` wrapper that always exists in the DOM (empty when DND is off).
- Add `dnd-response` / `dnd-response-am` wrapper templates in `dnd-toggle.html` that return the toggle partial plus an htmx OOB swap for `#dnd-indicator` in a single response.
- Handler renders the new response templates instead of the bare toggle partials.
- Navbar indicator now links to `/phones` (where the toggle lives) instead of `/settings#do-not-disturb` (the AM settings page has no DND section, so the old link landed on a nonexistent anchor).

### Intercom: DND on (chip appears in navbar via OOB swap)
intercom

### Dialup: DND on
dialup on

### Dialup: DND off (chip gone, toggle unchecked)
dialup off

### Answering Machine: DND on (red lamp appears in header)
am

## Test plan

- [x] `make test` passes
- [x] `make test-integration` passes
- [x] Visual verification: toggling DND on `/phones` updates both the toggle and the navbar indicator in all three themes
- [x] Visual verification: toggling DND off removes the navbar indicator via OOB swap
- [x] Settings page DND toggle (full POST redirect) still works unchanged
- [x] E2e chip selectors (`.rail .chip--err`, `.dialup-toolbar .chip--err`) still match through the new `<span>` wrapper

### DIFF
--- a/server/internal/web/handler_settings.go
+++ b/server/internal/web/handler_settings.go
@@ -113,7 +113,7 @@ func (h *Handler) handleSettingsDoNotDisturb(w http.ResponseWriter, r *http.Requ
 	if isHTMX(r) {
 		households[0].DoNotDisturb = enabled
 		data := h.buildLinesData(r, households[0], "")
-		renderWith(w, h.tmplPhones, partialFor(r, "dnd-toggle", "dnd-toggle-am"), data)
+		renderWith(w, h.tmplPhones, partialFor(r, "dnd-response", "dnd-response-am"), data)
 		return
 	}
 	http.Redirect(w, r, "/settings?saved=1", http.StatusSeeOther)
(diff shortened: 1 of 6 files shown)


justinlindh:
@claude review this PR

PR #409, opened 2026-05-04
fix(server): unpaired devices no longer show as online in UI

justinlindh:
## Summary

- Devices in pairing mode connect under the sentinel number `"unpaired"`. `Hub.OnlineNumbers()` already filtered this, but three call sites in the phone-detail handlers used `hub.Get(number) != nil` directly, bypassing the filter.
- Added `Hub.IsOnline(number string) bool` that explicitly returns `false` for the `"unpaired"` sentinel and otherwise delegates to `Get`.
- Updated `handlePhoneDetail`, `handlePhoneOnline`, and `handlePhoneEditGet` to call `hub.IsOnline(number)` instead of `hub.Get(number) != nil`.

## Root cause

After a device sends `TypeRepair` (triggered by the `*#0` service code), it stays registered in the hub under its old number until it disconnects and reconnects as `"unpaired"`. During that window the three direct `Get` calls would still return a non-nil connection and flag the line as online. The phones list and dashboard already went through `OnlineNumbers()` and were not affected.

## Test plan

- [ ] `make test` passes (all existing + 3 new `IsOnline` unit tests)
- [ ] `golangci-lint run ./...` reports 0 issues
- [ ] Manual: trigger `*#0` on a device, confirm the web UI shows the line as offline while the device reboots and reconnects in pairing mode

### DIFF
--- a/server/internal/signaling/hub.go
+++ b/server/internal/signaling/hub.go
@@ -292,6 +292,16 @@ func (h *Hub) LastSeenAt(number string) *time.Time {
 	return &t
 }
 
+// IsOnline returns true if number has an active hub connection and is not in
+// pairing mode. Unpaired devices connect under the sentinel "unpaired" number
+// and must not be presented as online in the UI.
+func (h *Hub) IsOnline(number string) bool {
+	if number == "unpaired" {
+		return false
+	}
+	return h.Get(number) != nil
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

PR #423, opened 2026-05-05
refactor(server): remove admind service

justinlindh:
## Summary

- Remove the `admind` admin dashboard service entirely
- Eliminates: the `admind` binary, its Dockerfile, the `internal/admin/` package (server, auth, DB, templates, static assets, stats client, tests), a dedicated Postgres database (`admin-db` in compose, `admindb` CNPG cluster in k8s), Helm chart templates, compose services, and the CI image-build step
- Keeps signald's `/internal/stats` endpoint and `ADMIN_SECRET` intact

## Why

admind was never fleshed out. It displayed 6 aggregate integers (users, households, phones, online devices, active calls, links) on a read-only page with no mutating operations, no user management, no device controls. The "Users" and "Households" pages were placeholder text. It hadn't received a feature commit since its initial creation.

It carried real cost for that: a separate container image on every release, a dedicated Postgres database just for its own login sessions, its own Tailwind build step, shared-secret wiring, and integration test infrastructure.

If there's ever a need for real admin tooling (user management, device banning, audit logs), it makes more sense to build something purposeful from scratch than to keep maintaining this skeleton.

## Test plan

- [x] `go build ./...` passes (signald still compiles)
- [x] `golangci-lint run ./...` reports 0 issues
- [x] `go test ./...` all tests pass
- [x] `helm lint charts/digits/` passes with defaults and all-features-enabled
- [x] `helm template` renders cleanly
- [ ] CI passes (server-ci, helm-ci, server-integration)


justinlindh:
@claude

justinlindh:
Addressed review feedback: emptied `test-db-init.sql` (no "was here" comments) and inlined the single-variant `Service` type into a package-level const. `metrics.New()` no longer takes a service parameter.

PR #424, opened 2026-05-05
refactor(server): drop unread Stats field from dashboard view model

justinlindh:
`dashboardData.Stats` / `dashStats` is computed on every dashboard render but no template binds to it. The AM partial reads `dashStatusVM.OnlineLines` / `.ActiveCalls`, and the intercom and dialup dashboard templates never reference any `dashStats` field.

Removed the unread struct. The two counters that the `Status` struct still needs (`OnlineLines`, `ActiveCalls`) are now built inline at the call site rather than via an intermediate `stats := dashStats{...}` that nothing reads.

Net diff: 17 lines deleted, 2 added.

### DIFF
--- a/server/internal/web/handler_dashboard.go
+++ b/server/internal/web/handler_dashboard.go
@@ -13,7 +13,6 @@ import (
 
 type dashboardData struct {
 	chromeData
-	Stats              dashStats
 	Lines              []lineRow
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude - I thought we used these for jitter in call? No? Can you confirm dead?

PR #444, opened 2026-05-06
feat(server): add PodDisruptionBudgets and topology spread for HA

justinlindh:
## Summary

- Adds PodDisruptionBudgets for signald and Redis Sentinel (maxUnavailable=1), preventing voluntary disruptions from killing all replicas simultaneously
- Adds `topologySpreadConstraints` to the signald Deployment (ScheduleAnyway, maxSkew=1) to actively balance pods across workers, supplementing the existing preferred anti-affinity
- Both PDBs and topology spread are gated on `replicas > 1`, so single-replica deployments are unaffected

## Context

Pre-drill preparation for chaos testing the cluster. Without PDBs, a `kubectl drain` or node maintenance could evict all signald pods at once. Without topology spread, the scheduler was placing 2 of 3 signald pods on the same worker.

## Test plan

- [ ] `helm template` renders PDBs and topology spread with multi-replica values
- [ ] `helm template` skips both with default single-replica values
- [ ] `helm lint` passes
- [ ] After merge and ArgoCD sync, verify `kubectl -n digits get pdb` shows both PDBs
- [ ] Verify signald pods are spread across all three workers

### DIFF
--- a/charts/digits/templates/deployment-signald.yaml
+++ b/charts/digits/templates/deployment-signald.yaml
@@ -16,6 +16,15 @@ spec:
         app.kubernetes.io/part-of: {{ include "digits.name" . }}
     spec:
       automountServiceAccountToken: false
+      {{- if gt (int .Values.signald.replicas) 1 }}
+      topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: kubernetes.io/hostname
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

PR #453, opened 2026-05-06
fix(server): report local connection count in device metric

justinlindh:
## Summary

- The `digits_signald_active_devices_current` gauge was calling `hub.OnlineNumbers()` which, with Redis cluster state enabled, queries all pods via Redis SCAN. Every replica reported the same cluster-wide total, making the "Devices per replica" Grafana panel show identical bars.
- Adds `Hub.LocalConnectionCount()` that counts only the local `h.conns` map (this pod's WebSocket connections).
- Uses `LocalConnectionCount()` for the Prometheus gauge so each pod reports its own connection count.

## Test plan

- [x] Unit tests pass (`make test`)
- [x] Lint clean (`golangci-lint run ./...`)
- [ ] After deploy, verify "Devices per replica" panel in Grafana shows different values per pod (based on which pod each device connected to)

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -143,7 +143,7 @@ func run(ctx context.Context) error {
 	// relay (signaling_errors_total). Live-state gauges sample the hub and
 	// tracker at scrape time so we never persist counts elsewhere.
 	mreg := metrics.New(version.Version, version.Commit)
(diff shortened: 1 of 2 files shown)


justinlindh:
Note: after this deploys, the homelab-k8s dashboard panels "Active devices" (id 3) and "Active devices and calls" (id 31) need to switch from `max()` to `sum()`, since each pod will then report only its local WebSocket count instead of the cluster-wide total.

PR #458, opened 2026-05-07
feat(firmware): phase-aware FSM LED, remove imperative LED management

justinlindh:
## Summary

- Firmware FSM consults the persisted phase byte when choosing idle LED pattern: UNPAIRED shows SLOW_PULSE, SETUP shows DOUBLE_PULSE, RECOVERY shows FAST_BLINK, PAIRED shows OFF
- Removes imperative LED:LOCK/LED:pattern management from digitsd setup and recovery mode entry
- Setup only uses LOCK during WiFi verification (CONNECTING pattern the firmware can't determine)
- resetPicoHardware sends UNLOCK instead of OFF so it clears stale locks without overriding the phase-based pattern
- Fixes pairing announcement goroutine leak on HOOK:ON while unpaired
- Removes explicit sp.LED("ON") on unpaired HOOK:OFF (firmware FSM DIAL_TONE state handles it)

## Test plan

- [ ] Boot unpaired: LED shows SLOW_PULSE immediately (firmware boot) and continues after digitsd starts
- [ ] Pick up handset: LED goes solid ON (firmware FSM enters DIAL_TONE)
- [ ] Hang up: LED returns to SLOW_PULSE (firmware FSM enters IDLE, phase-aware)
- [ ] Setup mode: LED shows DOUBLE_PULSE without digitsd sending explicit commands
- [ ] Setup WiFi verify: LED switches to CONNECTING (LOCK), returns to DOUBLE_PULSE on failure (UNLOCK)
- [ ] Recovery mode: LED shows FAST_BLINK without digitsd sending explicit commands
- [ ] Paired idle: LED stays OFF (no regression)

### DIFF
--- a/firmware/src/phone_fsm.c
+++ b/firmware/src/phone_fsm.c
@@ -14,12 +14,23 @@
 #include "tone.h"
 #include "uart_proto.h"
 
-// fsm_led_set wraps led_set_mode: skips FSM-driven LED changes when
-// the Pi has locked the LED (e.g. recovery/setup modes).
+// fsm_led_set resolves the LED pattern from (phase, fsm_state). When the Pi
+// has locked the LED (e.g. CONNECTING during WiFi verify), FSM changes are
+// skipped. When the FSM requests OFF (idle/on-hook), the phase determines
+// the idle pattern so each mode gets the right ambient indicator.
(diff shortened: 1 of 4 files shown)


justinlindh:
@claude

justinlindh:
@claude

PR #463, opened 2026-05-07
fix(firmware,pi,digitsd): boot state bugs (panic-check, hook init, pairing tone)

justinlindh:
## Summary

Three connected boot-state bugs:

- **Panic-check fails at boot**: the Pico isn't ready when `digits-panic-check` runs at sysinit. Added PING/PONG retry loop (5 attempts, 500ms spacing) before KEYDUMP so it waits for the Pico. Refactored serial reading to use a shared scanner goroutine to avoid data races between PING and KEYDUMP reads.
- **Firmware FSM ignores initial hook state**: if the handset is off-hook at power-on, no HOOK:OFF event fires because the FSM only reacts to state changes. Added `hook_is_off_hook()` check in `phone_fsm_init()` to emit HOOK:OFF over UART if the handset is already lifted.
- **Post-pairing dial tone dies on restart**: `mixer.PlayOnce("tone_dial")` dies when `os.Exit(0)` kills the process. Switched to `sp.SendFire("TONE:DIAL")` so the Pico plays the tone independently and it survives the digitsd restart.

## Test plan

- [ ] Hold `*` key on keypad during power-on: device enters recovery mode
- [ ] Power on with handset off-hook: digitsd log shows `RX line=HOOK:OFF` early, controller enters dial tone
- [ ] Pair device: Pico dial tone plays continuously through digitsd restart
- [ ] Normal boot (no keys held, handset on-hook): no regression, boots normally
- [ ] Panic-check logs: `PING: PONG` followed by `KEYDUMP` parsed successfully
- [ ] `go test -v` in `pi/digits-panic-check/`: 14/14 tests pass

### DIFF
--- a/firmware/src/phone_fsm.c
+++ b/firmware/src/phone_fsm.c
@@ -428,6 +428,9 @@ void phone_fsm_init(void) {
     s_state = PHONE_STATE_IDLE;
     clear_dialing_buffer();
     set_state(PHONE_STATE_IDLE);
+    if (hook_is_off_hook()) {
+        uart_proto_send("HOOK:OFF");
+    }
 }
 
 void phone_fsm_update(void) {
--- a/pi/digits-panic-check/main.go
+++ b/pi/digits-panic-check/main.go
@@ -55,6 +55,12 @@ func defaultConfig() config {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


justinlindh:
@claude

PR #500, opened 2026-05-08
fix(firmware,digitsd): star-at-boot recovery via persistent phase byte

justinlindh:
## Summary

Fixes the star-at-boot recovery trigger. Previously, pressing * during power-on sent a transient UART message (BOOT:PANIC) that was always lost because the Pi boots 15-30 seconds after the Pico. Now the Pico persists its intent to flash and digitsd reads it reliably.

- Firmware: write PHASE_RECOVERY to flash when * is held at boot, using debounce-free `keypad_scan_raw()` (the normal debounced scan rejected the key because <80ms elapsed since init)
- Firmware: add `PHASE?` UART command that returns the persisted phase byte
- digitsd: query phase after POST with retries; if RECOVERY, write flag file + boot counter, clear phase back to PAIRED/UNPAIRED, reboot into recovery partition
- digitsd: `doReboot()` now uses `sudo /usr/sbin/reboot` (in the digits user's sudoers allowlist) instead of bare `systemctl reboot` which lacked permissions
- Pin `Dockerfile.builder` to `golang:1.26-bookworm` (floating tag moved to Trixie, causing glibc 2.40 mismatch with Bookworm devices)

### Dev-mode fixes (from testing)

- Phase display: human-readable label ("paired") instead of raw hex
- Journal logs: use `sudo journalctl` with matching sudoers entry
- Log tab switching: clear cached content so UI updates on tab switch
- Stale `digits-recovery` comment fixed in initramfs

### Tested on device

1. Pico detects * at boot, writes PHASE_RECOVERY, LED fast-blinks
2. digitsd queries PHASE?, sees 0x04, writes recovery flags, clears phase, reboots
3. Initramfs sees flags, boots recovery partition
4. Recovery web UI + AP functional

### Root cause chain

The original BOOT:PANIC approach had three bugs stacked:
1. **Timing**: UART message sent before Pi opened serial port (15-30s gap)
2. **Debounce**: `keypad_scan()` rejected the held key because <80ms since `keypad_init()`
3. **Permissions**: `doReboot()` used `systemctl reboot` without sudo

## Test plan

- [x] Firmware builds cleanly
- [x] digitsd builds cleanly (Bookworm-pinned builder)
- [x] Phase query works (saw PHASE:0x01 in logs)
- [x] Star-at-boot triggers recovery (saw PHASE:0x04, recovery flags written, reboot into recovery)
- [x] Dev-mode web UI shows correct phase, journal logs, tab switching
- [ ] Recovery audio (pre-existing issue, separate from this fix)

### DIFF
--- a/firmware/src/keypad.c
+++ b/firmware/src/keypad.c
@@ -30,7 +30,7 @@ void keypad_init(void) {
     s_last_press_time = get_absolute_time();
 }
 
-char keypad_scan(void) {
+char keypad_scan_raw(void) {
     char pressed = '\0';
     const uint num_cols = board->keypad_num_cols;
 
@@ -46,6 +46,12 @@ char keypad_scan(void) {
 
         gpio_put(board->keypad_rows[row], 1);
     }
+
+    return pressed;
+}
+
+char keypad_scan(void) {
+    char pressed = keypad_scan_raw();
 
     absolute_time_t now = get_absolute_time();
     if (pressed != '\0' && pressed != s_last_key) {
--- a/firmware/src/keypad.h
+++ b/firmware/src/keypad.h
@@ -12,7 +12,11 @@
 void keypad_init(void);
 
 // Returns pressed key ('0'-'9', '*', '#', and on V1 also 'A'-'D')
-// or '\0' if no new key press.
+// or '\0' if no new key press. Debounced.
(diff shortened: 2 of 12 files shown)


justinlindh:
@claude

PR #548, opened 2026-05-11
fix(server): stop call duration timer from ticking after call ends

justinlindh:
## Summary
- The live detail page for calls and conferences unconditionally started a JS timer counting up from the start time, even for ended calls. Visiting an old call showed an ever-growing bogus duration.
- When a call/conference has ended, the page now renders a static duration in the same `mm:ss` / `h:mm:ss` clock format as the live timer, so there is no visual jump when a live call ends and the page reloads.
- Added `DurationS` to `ConferenceLinkHealthResp` so the conference template has access to the final duration.

## Test plan
- [ ] Visit an ended call at `/call/live/{id}`: timer should show static final duration, not tick
- [ ] Visit an ended conference at `/conference/live/{uuid}`: same behavior
- [ ] Start a live call, watch the timer tick, end the call: page reloads and timer freezes at the correct final value in the same format

### DIFF
--- a/server/internal/web/handler.go
+++ b/server/internal/web/handler.go
@@ -57,6 +57,15 @@ func baseTemplateFuncs() template.FuncMap {
 			}
 			return fmt.Sprintf("%d:%02d", seconds/60, seconds%60)
 		},
+		"fmtDurationClock": func(seconds int) string {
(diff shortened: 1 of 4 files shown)


justinlindh:
@claude

PR #561, opened 2026-05-14
feat(digitsd): voicemail greeting playback and *97 custom recording

justinlindh:
## Summary

- Plays a default voice-cloned greeting before the beep on voicemail auto-answer, so callers hear a spoken prompt instead of jumping straight into the recording.
- Lets users record a custom outgoing greeting by dialing `*97` from any handset; subsequent voicemails play that recording instead of the default.
- Dialing `*99` deletes the custom greeting and reverts to the default.

## How it works

**Playback chain (auto-answer path).** When ring-timeout fires and voicemail is enabled, `VoicemailAutoAnswer` runs greeting playback off the daemon goroutine: try the user's recorded greeting first; on `os.ErrNotExist` fall back to the embedded default WAV (`pi/tones/voicemail_greeting.wav`). After the greeting drains, a short prompt beep plays and the recording pipeline takes over. The flow is greeting, then beep, then recording.

**Audio injection.** `audio.Pipeline.PlayGreetingSamples([]int16)` reuses the same beep-slot buffer the prompt beep uses; the caller computes drain time as `samples/48 kHz + 100 ms` so the next stage does not step on the greeting tail. The custom-greeting path decodes Opus frames into a single PCM buffer, then injects.

**Recording chain (`*97`).** Hook-off and `*97` from `DIALTONE` enters `StateVOICEMAIL_RECORD_GREETING`. A short prompt beep plus 400 ms tail keep the beep out of the recording. The pipeline starts (if not already up), `Store.BeginGreetingRecording` opens a temp `.frames` file, and a goroutine consumes mic frames, Opus-encodes them, and appends to the recorder. Recording terminates on `#`, hook-on, or 60 s max duration; `Finalize` atomically renames over any prior greeting, and the FSM resets to `DIALTONE` with dial tone re-armed. Hook-on during a `*97` recording preserves what was spoken so far (real answering-machine convention) rather than discarding it.

**Storage.** Custom greetings live next to messages in the per-line voicemail dir as `greeting.frames` (Opus-framed PCM, same format as captured messages). `Store.DeleteGreeting` is idempotent on missing file. The default WAV is mono 16-bit PCM at 44.1 kHz, POTS-filtered, and ships verbatim under `pi/tones/`.

**Standalone Opus encoder.** Greeting record uses `codec.NewEncoder` directly (mic, then encoder, then recorder) instead of routing through a WebRTC peer, since there is no remote party for a greeting. Same 60-second cap as the message recorder.

## Bundled bug fixes

These landed on the branch during the work. Each is its own commit; calling them out so review can isolate them.

- **`fix(digitsd): unmute mic on voicemail call screening pickup`** (commit 270f1592). Phase 2 carry-over. `VoicemailPickup` bridged WebRTC audio to the mixer but never lifted the mute set by `VoicemailAutoAnswer`, so the homeowner's "Hello?" never reached the caller after a screening pickup. Fixed by setting `pipeline.SetMuted(false)` in the pickup path.
- **`fix(digitsd): remove blanket download timeout from updater`** (commit f57f8f21). Unrelated to greetings; landed on this branch during development. The 120s client-level `http.Client.Timeout` was killing 37 MB digitsd binary downloads on Pi Zero 2W over wifi. Replaced with a bounded `context.WithTimeout(30 s)` on `CheckVersion` only; downloads now run without a client-level cap.
- **`fix(digitsd): move default voicemail greeting to pi/tones source tree`** (commit 6fdd09b4). The default greeting WAV was originally committed under `pi/digitsd/internal/assets/embed/data/tones/`, which is the build-output staging dir wiped by `make embed` before being repopulated from `pi/tones/`. CI release builds would have silently deleted the asset. Moved to `pi/tones/voicemail_greeting.wav` (canonical source); `make embed` now copies it like every other tone.
- **`refactor(digitsd): align voicemail logging with call setup convention`** (commit 447f16a0). Every voicemail slog line now includes `caller` / `peer` and sync-elapsed timing, matching `HangupCall` / `AnswerCall`. Makes voicemail flows greppable in the same way live-call flows already are.
- **`refactor(digitsd): simplify greeting playback dispatch and fix IsCallActive`** (commit 3891eba0). Post-simplify pass. Adds `StateVOICEMAIL_RECORD_GREETING` to `IsCallActive` (the prior fix in d806ed91 missed this state; docstring already promised it). Drops the `hasCustomGreeting` pre-check and the TOCTOU stat-then-open race against `*99` delete. Extracts a `greetingPlaybackDuration` helper. Drops a redundant per-frame `samples := make + copy` in the Opus decode loop.

## How to test

Two phones, voicemail enabled in `config.json` (`voicemail.enabled: true`, short `ring_timeout` for faster iteration), restart digitsd.

1. Call the phone from another handset and let it ring through the timeout. The default voice-cloned greeting plays first, then the beep, then the recording window opens.
2. Pick up the called phone; dial `*97`. After the prompt beep, speak a short greeting. Press `#` to finish (or hang up; the partial recording is preserved either way). FSM returns to dial tone.
3. Call again from another handset and let it ring through. The custom greeting now plays in place of the default.
4. From the called phone, dial `*99` to delete. Call again; the default greeting returns.
5. Verify hook-on during `*97` recording saves what was spoken rather than discarding it, and the FSM exits to `IDLE` cleanly.
6. Sanity: screening pickup during voicemail (lift handset while greeting/beep/record is running) now bridges audio in both directions (the Phase 2 unmute fix).


justinlindh:
@claude

justinlindh:
Closing in favor of a clean recreation. This branch's merge commit pulled unsquashed Phase 2 history (from voicemail-capture) into the greeting branch, which collided with PR #560's squashed result on main. Replacement PR coming with the same content on a linear branch from main.

PR #566, opened 2026-05-14
feat(server): per-line voicemail settings

justinlindh:
## Summary

Phase 4 of the voicemail rollout: per-line voicemail settings live on the server and push to digitsd over the existing line_settings channel, mirroring silent_mode/voice_style/auto_update.

- Five new fields under a `voicemail` sub-object inside the existing per-line `settings` JSONB column: enabled, ring_timeout_seconds (5-60), max_message_seconds (15-180), max_stored_messages (5-200), retrieval_code (regex `^[0-9*#]{2,6}$` and must contain `*` or `#`).
- No DDL migration: voicemail nests inside the existing v11 settings JSONB column. Defaults applied in `line.DefaultSettings`, clamped in `Normalize` as defense in depth.
- New `POST /phones/{number}/voicemail` handler; rejects out-of-range / bad retrieval codes with 400, persists, emits a `line_settings` push including the voicemail block.
- Digitsd-side receiver decodes the new sub-blob, converts seconds to `time.Duration`, mutates `cfg.Voicemail.*` under `d.mu`, calls `cfg.Save()` outside the lock. `Enabled`, `RingTimeout`, `RetrievalCode` take effect on the next inbound call; `MaxStoredMessages` and `MaxMessageDuration` take effect after the next daemon restart (documented inline).
- Wire format: `LineSettings.Voicemail *Voicemail` (pointer, omitempty on envelope, no omitempty on inner fields so an authoritative server push can carry explicit `enabled: false` and zero ints rather than have them silently dropped).
- /simplify pass folded the persist+push+log tail into a single `applyLineSettings` helper shared between `handlePhoneVoicemailPost` and `updateLineSetting`, and collapsed two open-coded `line.Voicemail` to `signaling.Voicemail` field copies onto a new `signaling.VoicemailFromLine`. No behavior change.

## Side fixes

Three hardware-verified bug fixes that landed on this branch during testing:

- **Busy tone overlay on *98 exit.** When *98 found no messages (or reached end-of-messages), the busy ack tone played on top of the re-armed dial tone loop. `voicemailExitToDialtoneAsync` now drains the mixer's one-shot queue before re-arming the loop. Three call sites fixed.
- **Greeting WAV sample rate.** The default voicemail greeting WAV was authored at 44.1 kHz but the pipeline injects it at 48 kHz, causing a chipmunk pitch shift. Resampled `voicemail_greeting.wav` to 48 kHz. Other 44.1 kHz tone assets (call_return, wifi, recovery) are unaffected because they play through the mixer/ALSA path which handles the rate difference.
- **Bell not ringing after daemon restart.** `SendFire` (fire-and-forget UART writes) had no pacing. During init, three commands fired within 1ms, overflowing the Pico's 32-byte RX FIFO and silently dropping STATE:SET:PAIRED and subsequent RING:START commands. Added 15ms post-write delay to SendFire. Hardware-verified: bell now rings reliably after restart.

## Test plan

- [x] `golangci-lint run ./...` clean from `server/`
- [x] `golangci-lint run ./...` clean from `pi/digitsd/`
- [x] `make test` from `server/`
- [x] `make test-integration` from `server/`
- [x] `make test` from `pi/digitsd/`
- [x] `make build` from `pi/digitsd/`
- [x] /simplify pass
- [x] Hardware verified: *98 busy tone, greeting playback, bell ringing after restart

Phase 5 (UI) follows.


justinlindh:
@claude

PR #569, opened 2026-05-15
feat(digitsd,server): default answering machine to enabled

justinlindh:
## Summary

Flips the default voicemail enabled state from false to true in both the digitsd config defaults and the server-side line settings defaults. New lines and fresh device configs now have the answering machine enabled out of the box.

## Test plan

- [x] Server unit tests updated and passing
- [x] digitsd config tests passing
- [x] golangci-lint clean both modules


justinlindh:
@claude

PR #579, opened 2026-05-15
docs: add voicemail to the README service codes and docs links

justinlindh:
## Summary

The voicemail docs are linked from `docs/README.md` but were missing from the root `README.md`. This adds voicemail to both places it belongs in the root README.

- Adds a Voicemail subsection to the Service Codes section: a `*97`/`*98`/`*99` table matching the existing Call Return subsection style, plus a sentence on the auto-answer and mic-mute behavior.
- Adds a `Voicemail` entry to the Documentation links list, pointing at `docs/voicemail.md`.

Documentation only; no code changes.

## Test plan

- [ ] README renders: the new Voicemail subsection table and the docs link display correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,16 @@ Hidden codes entered on the keypad while the handset is off-hook.
 
 These are dialed from dial tone (not prefixed with `*#`), matching the original 1990s POTS behavior.
 
+### Voicemail (*97 / *98 / *99)
+
+| Code | Action | Details |
+|------|--------|---------|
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the review. Addressed the heading/table order note: fixed in 9285ac74. The heading is now `*98 / *97 / *99`, matching the table order, which leads with `*98` (listen) as the primary action. Heading and table are now consistent with each other, the same as the Call Return section.

PR #586, opened 2026-05-15
fix(digitsd): play the greeting-recording beep through the earpiece

justinlindh:
## The bug

Dialing `*97` to record a custom voicemail greeting plays the spoken prompt "Record your greeting after the tone", but the user hears no beep before recording starts. Confirmed by hardware testing.

## Root cause

`PlayGreetingBeep` arms the synthesized beep for injection into the CAPTURE loop: the beep replaces real mic frames so it travels in the OUTBOUND audio stream. That is correct for `VoicemailAutoAnswer`, where a WebRTC caller is on the other end and must hear the beep.

But the `*97` greeting-recording flow has no WebRTC peer and no local monitoring, so a beep injected into the capture/outbound path reaches nobody. The user's earpiece is fed by the mixer, not the capture path. The spoken prompt is heard in `*97` only because it plays via the mixer.

## The fix

- `pipeline.go`: the beep-synthesis body of `PlayGreetingBeep` is extracted into a new exported `SynthGreetingBeep` that returns the PCM buffer. `PlayGreetingBeep` keeps its outbound-capture behavior byte-identical (same 1kHz sine, same amplitude, same fade envelope, same zero-length guard).
- `voicemail.go` `VoicemailRecordGreeting`: synthesizes the same beep and plays it through the mixer with `PlayOnceSamples`, then waits via `waitForOnceComplete` so the beep finishes before the recording goroutine starts and is not captured. The old 400ms sleep, which compensated for the capture-path routing, is removed.
- `VoicemailAutoAnswer` is unchanged: its beep must reach the WebRTC caller via the outbound path.

This is a hardware-testing bug fix. No new audio assets: the existing beep waveform is reused.

## Verification

- `make build` (Docker cross-compile): passes
- `go test ./...`: passes
- `golangci-lint run ./...`: 0 issues

### DIFF
--- a/pi/digitsd/cmd/digitsd/voicemail.go
+++ b/pi/digitsd/cmd/digitsd/voicemail.go
@@ -497,11 +497,16 @@ func (d *daemonCallbacks) VoicemailRecordGreeting() {
 	d.greetingRecorderMu.Unlock()
 
 	// Spoken prompt ("Record your greeting after the tone") followed by the
-	// beep so the user knows the recorder is hot. 400ms tail keeps the beep
-	// out of the recording itself.
+	// beep so the user knows the recorder is hot. The *97 flow has no WebRTC
+	// peer, so the beep is played through the mixer into the earpiece rather
+	// than injected into the outbound capture path (which nobody would hear).
+	// waitForOnceComplete gates the recording goroutine until the beep has
+	// finished, keeping it out of the recording itself.
 	d.playAnnouncementSequence("vm_record_greeting")
-	pipeline.PlayGreetingBeep(300 * time.Millisecond)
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the review. On the one observation: the hardcoded `48000` is intentional and matches the existing convention in `voicemail.go` (the encoder init and the rest of the file) and `DefaultPipelineConfig`. The mixer and all tone assets are 48kHz; the V2 codec resamples to 44.1kHz at the ALSA layer, so 48000 is the correct rate for anything fed to the mixer. If the rate ever becomes per-device configurable, every one of those sites moves together. No change.

PR #587, opened 2026-05-15
feat(digitsd): announce "Message N" before each voicemail during *98 playback

justinlindh:
## Summary

During `*98` voicemail retrieval, messages played back to back with no separator. This adds a spoken "Message one", "Message two", etc. before each message when the session holds two or more messages, so the listener can tell them apart. A lone message is left unannounced, since the "you have 1 message" count intro already identifies it.

- New voice asset `pi/tones/vm_message.wav` ("Message"), generated with the voice-clone tool using the same bell-woman reference voice as the existing voicemail prompts.
- `announceMessageNumber` composes `vm_message` + `spoken_N`, the same composition pattern as the "you have N messages" intro, reusing the existing `spoken_0` through `spoken_9` digit clips.
- The announcement plays at the start of `voicemailPlaybackLoop`. Whether to announce is decided once at session entry from the unheard count.
- The 1-based position rides on the playback session. `openNextUnheardLocked` bumps the counter for each new message; the `*` replay path keeps the message's existing number so a replay re-announces the same "Message N".
- A position above 9 has no digit clip, so the bare "message" word plays as a separator cue.
- `docs/voicemail.md` updated so the asset table and composition description match HEAD.

## Test plan

- [x] `make build` (Docker cross-compile) passes
- [x] full `pi/digitsd` test suite passes, including the new `messageNumberClips` unit test
- [x] `golangci-lint` clean on `cmd/digitsd`
- [ ] hardware: dial `*98` with 2+ messages, confirm "Message N" plays before each


justinlindh:
@claude

justinlindh:
Thanks for the review. Going through the observations:

1. **messageNumberClips and test.** No change needed.

2. **Announcement flag set after the count plays.** Acknowledged, and I agree it is not worth synchronizing. The window is a DTMF press during the ~2s count intro, and the only consequence is one message missing its "Message N" prefix. Adding a lock or a pre-armed flag to close a 2-second cosmetic window is not a good trade. Leaving as is.

3. **Context check before the announcement.** Intentionally consistent with announceMessageCount and the other blocking announcement calls. No change.

4. **Replay preserves position.** Confirmed intended behavior. No change.

5. **Struct field alignment in main.go.** The two new fields (`voicemailAnnounceNumbers`, `voicemailMessageSeq`) are already gofmt-aligned to each other; `gofmt -d` does not touch those lines. The hunks gofmt does want are in pre-existing field groups higher up in the struct: both `main.go` and `voicemail.go` are already gofmt-dirty on `main` before this PR (verified against `github/main`). Running `gofmt -w` here would reformat ~24 lines of unrelated struct fields and mix that churn into this feature PR, so I am leaving it out of scope. It is a pre-existing repo papercut, not introduced here.

PR #590, opened 2026-05-15
fix(digitsd): start voicemail recording after the greeting, not at connect

justinlindh:
## Summary

Hardware testing found a large delay during `*98` voicemail retrieval between a message starting and the caller's audio actually playing. The root cause is on the recording side, not playback.

In `VoicemailAutoAnswer`, the `OnRemoteTrack` handler reaches its live phase right after the call connects and from there teed every caller frame into the recorder. But the recorder is opened before the outgoing greeting plays, and a separate goroutine runs a 500ms lead-in, the greeting, a 500ms beep, and a 500ms tail before transitioning into the recording state. So every stored message was front-loaded with the entire greeting duration of the caller silently listening, followed by the actual message. On `*98` playback that silence is what plays between the message announcement and the caller's first words. With a custom greeting (up to 60s) it was also consuming most of the 90s message cap.

## Fix

- Add a `voicemailRecordArmed` atomic flag. The Phase 3 tee into the recorder is gated on it.
- The greeting goroutine sets it true at the exact point it transitions into the recording state, right where the mic is muted (after the beep). Recording now starts when the caller starts speaking.
- Frames the live track delivered before arming are still decoded, to keep the Opus decoder state and the pickup PCM channel in sync. They are simply not recorded.
- A caller who hangs up during the greeting now appends zero frames. `Recorder.Finalize` already discards a zero-frame recording, so that path stops saving junk silent voicemails. `HangupCall` now sets its `finalizedVoicemail` flag from the finalized message ID, so a discarded zero-frame recording no longer triggers a spurious message-waiting LED re-emit.

The fix mirrors the existing `SetMuted(true)` gating, which is already called at the same transition point.

## Test plan

- [x] `make build` (Docker cross-compile) passes
- [x] full `pi/digitsd` test suite passes (`go test ./...`)
- [x] `golangci-lint run` clean
- [ ] hardware: leave a voicemail, dial `*98`, confirm the recording starts at the caller's first words with no greeting-length silence

## Notes on testing

The bug lives in the timing between the `OnRemoteTrack` goroutine and the greeting goroutine, driven by a pion `webrtc.TrackRemote`. There is no fake-track harness in the daemon test suite and building one is disproportionate to a four-line gate. The recorder behavior the fix relies on, a zero-frame `Finalize` discarding the recording, is already covered by `TestEmptyRecordingDiscarded`. Hardware verification covers the integration.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -149,6 +149,15 @@ type daemonCallbacks struct {
 	recorderMu        sync.Mutex
 	voicemailWebRTCCh chan []int16
 
+	// voicemailRecordArmed gates the OnRemoteTrack tee into the recorder.
+	// The remote track reaches its live phase right after the call
+	// connects, well before the outgoing greeting and prompt beep finish,
+	// so appending from that point would prepend the whole greeting
+	// duration of caller-side silence to every message. The greeting
+	// goroutine sets this true at the moment it transitions into the
+	// recording state, so the recorder only ever sees the actual message.
+	voicemailRecordArmed atomic.Bool
+
 	// Greeting recording (separate from message recording above). Active
 	// only between *97 entry and either # / hook-on / max-duration. Lives
 	// under its own mutex so it doesn't contend with the message recorder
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the thorough review, the teardown-path table matches my analysis.

On the `HangupCall` observation: I checked `Recorder.Finalize` and it never returns a non-zero `Message` together with a non-nil error. Every error path (fsync, close, rename, writeMeta) returns a zero `Message{}` with the error; the only non-zero-ID return is the final success path with a nil error. So `finalizedVoicemail = msg.ID != 0` is correct in practice: on any error `msg.ID` is 0 and the flag is false. The error-path note is accurate as a general caution but cannot misfire against the current Finalize contract. No change.

PR #592, opened 2026-05-15
feat(digitsd): review saved voicemails and speak playback controls during *98

justinlindh:
## Summary

Hardware testing surfaced two gaps in `*98` voicemail retrieval: a message played once was marked heard and became permanently unreachable from the handset (it felt deleted after the first listen), and the DTMF playback controls existed with nothing telling the user about them. This adds a saved-message review path and a spoken controls prompt.

### Saved-message review

- At `*98` entry the session snapshots the IDs of every already-heard message as a saved-review queue, taken before playback marks anything heard.
- Unheard messages play first, exactly as before, including the "Message N" announcements.
- When the unheard messages run out, `*98` announces the saved-message count and plays the heard messages in order, instead of ending.
- If there are no unheard messages but saved ones exist, `*98` enters the saved review directly rather than reporting an empty mailbox. A truly empty mailbox keeps the "no messages" behavior.
- Auto-mark-heard-on-EOF and the unheard-count LED are unchanged. This adds a review path; it does not change what "new" means.

### Spoken playback controls

- A new `vm_playback_controls.wav` prompt plays once per `*98` session, after the first count announcement, so the controls are discoverable.
- It says: "While a message is playing, press seven to delete it, pound for the next message, or star to replay it."

### Key semantics

- `7`: delete the current message, advance. Works in both the new and saved phases.
- `9`: in the new phase, mark the current message heard and advance. It is the only way to mark a message heard without listening to the whole thing (`#` leaves it unheard, playing to the end auto-marks it). In saved review the message is already heard, so `9` just advances. `9` is intentionally not announced, since the default behavior already saves a message.
- `#`: skip to the next message, leaving the heard flag untouched.
- `*`: replay the current message from the start.

### New voice assets

`vm_saved_message.wav` and `vm_saved_messages.wav` compose with the existing "you have" and digit clips to announce the saved count; `vm_playback_controls.wav` speaks the controls. All voice-cloned with the standard voicemail voice and user-approved. Ten or more saved messages reuse the existing `vm_lost_count` phrase.

## Test plan

- [x] `make build` (Docker cross-compile) passes with the new assets embedded
- [x] full `pi/digitsd` test suite passes, including a new `TestSavedCountClips`
- [x] `golangci-lint run` clean
- [ ] hardware: `*98` plays new messages, then announces the saved count and plays saved messages; the controls prompt is heard once; `7` deletes a message and its files are gone


justinlindh:
@claude

justinlindh:
Thanks for the thorough review, especially tracing the three saved-review entry points and the concurrency edges.

On the one observation: I went ahead and added the two-line cleanup. `VoicemailExitPlayback` now nils `voicemailSavedQueue` and resets `voicemailSavedCursor` under `voicemailMu`. It is still not a correctness fix (`snapshotSavedQueueLocked` reinitializes both at the next entry), but it makes the exit path symmetric with how it already releases the session and mixer source, and it keeps a stale slice from lingering between sessions. Done in commit 1df7ef43.

Both test phones are redeployed to the PR tip (1.37.1-4-g1df7ef43). The hardware test items remain the gate.

PR #595, opened 2026-05-16
fix(digitsd): stop the spurious WARN when *97 finalize races the recorder

justinlindh:
## Summary

Ending a `*97` greeting recording (the `#` key, hang-up, or the 60s duration cap) calls `finalizeGreetingRecording`, which closes the recorder. The mic-to-Opus-to-recorder goroutine can already be mid-iteration holding the recorder pointer it read a moment earlier, so its next `AppendFrame` lands on the closed recorder and the loop logged:

```
WARN voicemail: greeting append failed error="voicemail: recorder closed"
```

The greeting is finalized and saved before that stray frame runs (`custom greeting saved` logs immediately before the WARN), so it is purely cosmetic, but it printed once per greeting recording. Found by log-watch during `*96`/`*97` hardware testing.

## Change

- `internal/voicemail/store.go`: promote the bare `errors.New("voicemail: recorder closed")` to a package sentinel, `voicemail.ErrRecorderClosed`, returned from `Recorder.AppendFrame`.
- `cmd/digitsd/voicemail.go`: the greeting-record loop treats `errors.Is(err, voicemail.ErrRecorderClosed)` as a clean stop and returns quietly. Every other `AppendFrame` error keeps the existing WARN-and-continue.

Pre-existing behavior, not introduced this session and unrelated to `*96`.

## Test plan

- [x] `go test ./...` and `golangci-lint run ./...` pass
- [ ] Hardware: record a `*97` greeting, press `#`, confirm `custom greeting saved` still logs and the `recorder closed` WARN no longer appears

### DIFF
--- a/pi/digitsd/cmd/digitsd/voicemail.go
+++ b/pi/digitsd/cmd/digitsd/voicemail.go
@@ -604,6 +604,13 @@ func (d *daemonCallbacks) VoicemailRecordGreeting() {
 
 			atCap, err := rec.AppendFrame(payload)
 			if err != nil {
+				if errors.Is(err, voicemail.ErrRecorderClosed) {
+					// finalizeGreetingRecording closed the recorder out from
+					// under this loop (the # key, hang-up, or duration cap).
+					// The greeting is already saved; this frame just lost the
+					// race. Stop quietly instead of logging a spurious WARN.
+					return
+				}
 				slog.Warn("voicemail: greeting append failed", "error", err)
 				continue
 			}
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude please review this PR

justinlindh:
Added commit 77e12c46: folds in the message-record-path guard you flagged.

One correction to the suggested snippet. On the message path the guard suppresses the WARN but does **not** `return`. Unlike the `*97` greeting loop, the `OnRemoteTrack` goroutine has a second job after the recorder closes: it keeps decoding the caller's audio into `webrtcCh`, which `VoicemailPickup` hands to the mixer as the live call's audio. Returning on `ErrRecorderClosed` would kill live-call audio the instant the homeowner picks up. So the message-path fix is `if !errors.Is(err, voicemail.ErrRecorderClosed) { slog.Warn(...) }` with the loop falling through, while the greeting loop keeps its `return` (no live call follows there).

@claude please review the new commit.

PR #596, opened 2026-05-16
refactor(server): remove the per-line voicemail max-message-duration setting

justinlindh:
## What

Removes the configurable per-line voicemail max-message-duration setting. Per owner direction, that knob was pointless configurability. The per-message recording cap becomes a fixed 10-minute limit hardcoded in digitsd, so the server no longer stores, validates, or pushes a per-line value.

This is the server side. The digitsd side (fixed `10 * time.Minute` constant, dropping the wire field) is handled separately.

## Changes

- `line.Voicemail`: drop `MaxMessageSeconds` and its constants (`DefaultVoicemailMaxMessageSeconds`, `VoicemailMaxMessageMin`, `VoicemailMaxMessageMax`) plus the `Normalize()` clamp and `Merge()` handling.
- `signaling.Voicemail` wire struct: drop the `max_message_seconds` key so the `line_settings` push no longer carries it. digitsd stops expecting it.
- Phone detail voicemail form (intercom and answering-machine templates): remove the "Max message" input. The enabled toggle, ring timeout, max stored messages, and retrieval code are unchanged.
- Voicemail settings handler (`/phones/{number}/voicemail`): drop the `max_message_seconds` parse and range validation.
- Tests updated: dropped `TestPhoneVoicemailMaxMessageOutOfRangeRejected`, adjusted the JSON round-trip and clamp tests, and added regression coverage that the wire and new JSONB writes never carry `max_message_seconds`.

## JSONB compatibility

`lines.settings` rows written before this change still carry `max_message_seconds`. `encoding/json` drops unknown keys silently, so legacy rows decode cleanly. `TestSettingsUnmarshalIgnoresVestigialMaxMessageKey` pins that behavior. No migration needed: the key is simply vestigial and gets dropped on the next write.

## Wire contract

Coordinated with the digitsd side: server stops emitting `max_message_seconds`, digitsd stops expecting it and hardcodes the 600s cap. Remaining voicemail wire fields are unchanged: `enabled`, `ring_timeout_seconds`, `max_stored_messages`, `retrieval_code`.

## Testing

- `go build ./...` clean
- `go test ./...` passes
- `golangci-lint run ./...`: 0 issues

### DIFF
--- a/server/e2e/tests/10-voicemail.spec.ts
+++ b/server/e2e/tests/10-voicemail.spec.ts
@@ -77,12 +77,12 @@ test.describe('Voicemail (intercom theme)', () => {
     await expect(page.locator('#voicemail-section input[name="enabled"]')).toBeVisible();
     await expect(page.locator('#voicemail-section details.voicemail-advanced')).toBeVisible();
 
-    // Advanced should be collapsed by default; the 4 fields are hidden until expanded.
+    // Advanced should be collapsed by default; the detail fields are hidden until expanded.
     const ringField = page.locator('#voicemail-section input[name="ring_timeout_seconds"]');
     await expect(ringField).not.toBeVisible();
   });
 
-  test('expanding Advanced reveals all four detail fields', async ({ page }) => {
+  test('expanding Advanced reveals all detail fields', async ({ page }) => {
     const href = await firstPhoneHref(page);
     if (!href) {
       test.skip(true, 'No phones registered');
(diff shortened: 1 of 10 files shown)


justinlindh:
@claude please review

justinlindh:
Thanks for the review. No changes needed: the removal is server/-only and the JSONB backward-compat is covered by `TestSettingsUnmarshalIgnoresVestigialMaxMessageKey`. Leaving the PR for the team to merge.

PR #597, opened 2026-05-16
feat(digitsd): fix the voicemail message cap at 10 minutes, drop the setting

justinlindh:
## Summary

The per-line max-message-duration voicemail setting is removed. It was pointless configurability: a message ends when the caller hangs up, and the duration cap is only a backstop for a caller who never does. A fixed, generous cap covers that without a knob nobody needs to turn.

- `internal/voicemail/store.go`: new fixed package constant `messageMaxDuration` = 10 minutes; the `Options.MaxMessageDuration` field is gone and `BeginRecording` caps the recorder at the constant, mirroring how greeting recording already uses `greetingMaxDuration`.
- `internal/config/config.go`: the `MaxMessageDuration` field is removed from the `Voicemail` config struct and from `defaultVoicemail()`.
- `internal/signal/protocol.go`: `max_message_seconds` is dropped from the wire voicemail-settings struct.
- `cmd/digitsd/main.go`: the `line_settings` receiver no longer reads or applies a max-message value; the voicemail store opens with only `MaxStoredMessages`.
- `cmd/digitsd/voicemail.go`: when an inbound recording hits the 10-minute cap, the caller now hears the prompt beep before the call is hung up, so a runaway recording is not dropped into silent dead air.
- Docs: `voicemail.md` and `voicemail-guide.md` updated for the fixed 10-minute cap and the cap beep; the configurable 15-180s language is removed.

Recording-ends-on-caller-hangup is unchanged.

## Coordination

The wire-contract change (dropping `max_message_seconds` from the `line_settings` voicemail payload) is paired with the server-side removal in task #22. `encoding/json` ignores unknown fields, so the two sides need no flag-day and can land in either order.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...`, `golangci-lint run ./...` all pass
- [x] `TestMessageMaxDurationCap` reworked to the fixed constant
- [ ] Hardware: leave a normal voicemail, hang up, confirm it saves and ends cleanly (unchanged path)
- [ ] Hardware: a recording that runs to the cap plays a beep to the caller before the call drops

### DIFF
--- a/docs/voicemail-guide.md
+++ b/docs/voicemail-guide.md
@@ -103,13 +103,12 @@ the phone simply rings and never auto-answers.
 
 ### The settings form
 
-Open the advanced settings block in the voicemail section to edit four fields.
+Open the advanced settings block in the voicemail section to edit three fields.
 Change what you need, then submit the form to save.
 
 | Field | What it controls | Allowed range |
 |-------|------------------|---------------|
 | Ring timeout | How many seconds a call rings before the answering machine picks up | 5 to 60 seconds |
-| Max message | The longest a single caller message can be | 15 to 180 seconds |
 | Max stored messages | How many messages the phone keeps before the oldest is dropped | 5 to 200 |
 | Retrieval code | The code you dial on the phone to listen to messages | 2 to 6 characters, digits and `*` and `#` only, must include at least one `*` or `#` |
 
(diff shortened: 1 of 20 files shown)


justinlindh:
@claude

justinlindh:
Closing this PR. Its branch picked up an unrelated server-side commit through a shared-worktree collision (go-backend's `git add -A` for task #22 swept files in), so the diff here is contaminated with server changes that belong to PR #596.

The clean digitsd-only replacement is **PR #598**, a single commit off `main` with the full task #21 work plus the recovered message-record-path WARN guard. Nothing is lost.

PR #602, opened 2026-05-16
fix(digitsd): differentiate new vs saved messages in *98 retrieval

justinlindh:
## Summary

- When dialing *98 with 0 unheard but saved messages on disk, the phone played a confusing `vm_lost_count` clip ("you have so many messages I lost count") because the saved count (10) exceeded the 9-clip limit. The user had no way to know these were old/saved messages.
- Cap spoken message counts at 9 for both new and saved announcements instead of falling back to the ambiguous `vm_lost_count` clip.
- On direct entry to saved review (0 unheard, saved exist), announce "You have no new messages" first so the user knows nothing is urgent before saved messages start playing.
- When transitioning from new to saved phase, announce "Saved messages are next" before the saved count so the user knows the context changed.
- Add two new generated voice clips: `vm_no_new_messages.wav` and `vm_saved_next.wav`.

## Test plan

- [ ] Unit tests pass (`make test` in pi/digitsd)
- [ ] golangci-lint clean
- [ ] Deploy to 192.168.2.28 and dial *98 with 0 unheard / 10 saved: hear "no new messages", then "you have 9 saved messages", then controls, then saved playback
- [ ] Verify new-message path still works: leave a voicemail, dial *98, hear "you have 1 new message"


justinlindh:
@claude

PR #608, opened 2026-05-18
refactor(server): named constants, OneEnv helper, and minor cleanup

justinlindh:
## Code quality audit: server/

Before grade: **84 / 100**
After grade: **90 / 100**

---

### What was graded

The audit covered all Go source under `server/` for idiomatic style, naming, test coverage, project layout, and dead/stale patterns. Build and unit tests were clean throughout (`go build ./...`, `go vet ./...`, `go test ./...`). No high-severity issues were found; the three areas below accounted for the gap between 84 and the ceiling.

---

### Issues addressed

**1. Config env-parsing inconsistency (-3)**

`SIGNALD_LINK_HEALTH_FLUSH_DISABLED` and `TEST_FAKE_UPDATES` were parsed with inline `os.Getenv() == "1"` blocks instead of a helper, while every other env var used `StringEnv`, `BoolEnv`, or `IntEnv`. The reason for `"1"` (matching the Pi daemon's convention) was documented only on the first variable.

Fix: add `OneEnv` to `config/env.go` alongside the existing helpers, use it for both vars, and add tests to `config_test.go`. The `"1"` convention is now captured in the function doc once rather than in ad-hoc comments.

**2. Magic numbers (-5)**

Several numeric and string literals had no names, making their purpose and relationships non-obvious:

| Location | Literal | Now |
|---|---|---|
| `cmd/signald/main.go` | `300` (release cache seconds) | `releaseCacheTTL = 5 * 60` |
| `internal/signaling/relay.go` | `30 * time.Minute` (*69 retry window) | `callReturnExpiry` |
| `internal/signaling/relay.go` | `"stun:stun.l.google.com:19302"` | `googleSTUN` |
| `internal/web/handler_ws.go` | `10 * time.Second` (register deadline) | `wsRegisterTimeout` |
| `internal/web/handler_ws.go` | `32` (send channel buffer) | `wsSendBuf` |

The `ws*` constants in `handler_ws.go` were also promoted from a mid-function `const` block to package level so `wsRegisterTimeout` could be referenced before the block where it was previously declared.

**3. Trailing blank line in `device/store.go` (-1)**

Single spurious blank line at end of file; removed.

---

### Why the remaining 10 points are not addressed

- **Exported fields on `Relay`** (`Hub`, `Tracker`, `TURNGen`, etc.) are set from `cmd/signald/main.go` at startup and need to be exported for that; changing this would require a larger constructor refactor.
- **`handler.go` at 776 lines**: the file is organized coherently (template parsing, websocket upgrader, rate limiters, router) and splitting it would not improve clarity without a more significant restructure.
- **Template-function thresholds** (`pctToSegments`, `msToSegments`, `humanBytes`): the numeric thresholds are domain-specific quality boundaries defined in the same file as the rendering logic. Naming them would add noise rather than clarity at this scale.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KdJaQAGAdupe1Ry6QUmywu)_

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -36,6 +36,9 @@ import (
 	"github.com/justinlindh/digits/server/internal/web"
 )
 
+// releaseCacheTTL is the release index cache lifetime, in seconds.
+const releaseCacheTTL = 5 * 60
+
 func main() {
 	logging.Setup()
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
@@ -254,7 +257,7 @@ func run(ctx context.Context) error {
 	case cfg.GitHubRepo != "":
 		parts := strings.SplitN(cfg.GitHubRepo, "/", 2)
 		if len(parts) == 2 {
-			handler.Releases = updates.NewGitHubReleases(ctx, parts[0], parts[1], cfg.GitHubToken, 300, // 5 min cache
+			handler.Releases = updates.NewGitHubReleases(ctx, parts[0], parts[1], cfg.GitHubToken, releaseCacheTTL,
 				func(piLatest, fwLatest string) {
 					slog.Info("updates: new release detected, broadcasting", "pi", piLatest, "fw", fwLatest)
 					handler.Hub().Broadcast(&signaling.Message{
--- a/server/internal/config/config.go
+++ b/server/internal/config/config.go
@@ -3,8 +3,6 @@
 // environment values and built-in defaults.
 package config
 
-import "os"
-
 // Config holds all runtime configuration for signald, populated by Load from
 // environment variables. Zero values are overridden by Load's defaults where
 // appropriate.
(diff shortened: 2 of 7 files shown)


justinlindh:
@claude - I just a STUN server for digits, but this change references Google STUN. Does this break that?

justinlindh:
No behavior change here — `googleSTUN` is a named constant for the string `"stun:stun.l.google.com:19302"` that was already hardcoded in `handleRequestICE` before this PR. The rename is purely cosmetic.

Looking at the current logic: Google STUN is always included as the first ICE server entry, and if `SIGNALD_TURN_DOMAIN` is configured your TURN server is appended as a second entry. TURN servers also answer STUN requests, so devices that can reach your TURN server effectively have redundant STUN coverage. Nothing in this PR adds, removes, or reorders those entries.

If you want to replace the Google STUN entry with your own STUN server, that would be a separate change to `handleRequestICE` (and likely a new config field) — out of scope for this cleanup PR.

---

PR #632, opened 2026-05-24
fix(hardware): finalize V3 PCB for first fab

justinlindh:
## Summary
- Rename the Pi Zero W header reference to `PI_ZERO_W1`. The old `Pi Zero W1` contained spaces, which JLCPCB splits into separate designators, so the BOM and CPL never matched and the 2x20 header silently dropped from assembly. The same part (`C2977589`) assembled fine in V2, where it was named `J1`.
- Correct R13 to the real 120R part `C25079` (was a 12k `C25752`), fixing the +3V3 indicator LED current limit.
- Bring the remaining pre-fab board corrections into main: boost `SW_NODE` widened to 0.75mm with a hard min-width DRC rule, exposed-pad ground vias restored on the codec, driver, and RP2040, the Pi header mirrored to the correct mating handedness, and the SW1 DPDT pin map matched to the IL2240 (end-terminal common, multimeter-verified).

## Test plan
- [x] `kicad-cli pcb drc --schematic-parity` clean: 0 violations, 0 unconnected, 0 parity (verified locally)
- [ ] JLCPCB BOM/CPL upload matches `PI_ZERO_W1` and places the 2x20 header
- [ ] Bench bring-up of the fabbed board: boost ~37V, bell, mic-kill, hook sense


justinlindh:
@claude

PR #675, opened 2026-06-02
fix(digitsd): keep Pico forwarding keypad DTMF during voicemail sessions

justinlindh:
## Problem

On a phone with voicemail, dialing `*98` to retrieve messages plays the prompts and messages fine, but pressing keys to delete (`7`), save (`9`), skip (`#`), or replay (`*`) does nothing. The same applies to `*97` greeting record, where `#` is supposed to end the recording. This effectively never worked for a normal session.

## Root cause

Voicemail retrieval (`*98`) and greeting record (`*97`) run as **peerless, digitsd-local** sessions: there is no WebRTC peer, and digitsd handles the keypad menu itself.

The Pico firmware only forwards keypad events over UART while its FSM is in `DIALING` or `CONNECTED` (`firmware/src/phone_fsm.c:416`). Dialing a service code like `*98` leaves a partial dial buffer (`"98"`, the `*` is not a digit) that never reaches the 7-digit threshold to fire `DIAL:`. So the Pico's 15s inter-digit timeout (`DIAL_TIMEOUT_MS`, `phone_fsm.c:526`) fires and flips it to `BUSY`, after which **every keypress is silently dropped**. Since voicemail playback runs well past 15s, by the time the user presses a key the Pico is already `BUSY`.

Device logs made this unambiguous: across a ~63s `*98` session there was not a single `RX line=KEY:` event, even though plain dialing logs keys fine.

## Fix

digitsd-side only, no firmware reflash. The firmware already does the right thing on `CALL:CONNECTED` (accepts the transition from `DIALING` into `CONNECTED`, which both opens the key gate and stops the partial-dial timeout); it just was never told to do so for voicemail.

- On entering `*98` playback / `*97` record, send `CALL:CONNECTED` (the same command used when a real call answers) so the Pico stays in a key-forwarding state for the whole session.
- On returning to dial tone, send a Pico `RESET` so it leaves `CONNECTED` and can collect a fresh dial buffer again (normal calls fire `DIAL:` only from `DIALING`). This is gated on the voicemail interactive states inside `ResetToDialtone`, so it covers every exit path (end of messages, `#`, duration cap, error paths) in one place.
- The hangup path is unaffected: on `HOOK:ON` the Pico self-resets to `IDLE` and `ResetToDialtone` is not on that path.

The Pico shows its `CONNECTED` breathing LED during the session, which reads naturally as an active "call to your mailbox."

## Tests

- New controller unit tests: hold fires on `*98`/`*97` entry, release fires on exit to dial tone, and a non-voicemail `ResetToDialtone` does **not** poke the Pico.
- Full `pi/digitsd` suite passes; `golangci-lint` clean.

## Verification status

On-device verification is pending: the test device dropped off the network mid-session. Once it is back I will deploy this build and confirm, with the handset, that `7`/`9`/`#`/`*` now land during playback (logs show `RX line=KEY:` past the 15s mark and the matching delete/save/skip action), and that dialing still works after a voicemail session ends without hanging up.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -312,6 +312,10 @@ func (d *daemonCallbacks) NotifyCallConnected() {
 	d.serial.CallConnected()
 }
 
+func (d *daemonCallbacks) NotifyPicoReset() {
+	d.serial.Reset()
+}
+
 func (d *daemonCallbacks) EnableFlashDetection() {
 	d.serial.EnableFlashDetection()
 }
--- a/pi/digitsd/internal/phone/controller.go
+++ b/pi/digitsd/internal/phone/controller.go
@@ -94,7 +94,8 @@ type Callbacks interface {
 	InitiateCall(number string) error // Start outgoing WebRTC call
 	AnswerCall()                // Accept incoming WebRTC call
 	HangupCall()                // Tear down WebRTC call
-	NotifyCallConnected()       // Notify the Pico that the WebRTC peer answered
+	NotifyCallConnected()       // Notify the Pico that the WebRTC peer answered (Pico -> CONNECTED)
+	NotifyPicoReset()           // Reset the Pico FSM to IDLE, releasing a CALL:CONNECTED hold taken for a peerless local session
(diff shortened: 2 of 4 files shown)


justinlindh:
@claude

PR #680, opened 2026-06-02
fix(server): end grace-expired calls like a hangup so call-return and persistence fire

justinlindh:
## Summary

- Grace-window expiry now tears a 2-party call down exactly like an explicit hangup. Previously the grace timer called `ClearByNumber` (which fires `OnCallEndedNotify` with an empty callee) plus a single peer hangup; the normal hangup path calls `OnCallEnded(from, peer)` per peer, which persists the call end and fires the call-return observer with both names so `*69` retries targeting the peer run.
- Extracts a shared `endActiveCallsAsHangup` helper used by both `handleHangup` and the grace-expiry callback, so the two teardown paths cannot drift again. The idle and conference disconnect paths keep using `ClearByNumber` (no active 2-party call to end there).

## Why

Follow-up to a code-review finding on #678 (the reconnect grace window). On the merged code, a grace-expired call cleared in-memory and DB state but did not fire the call-return observer with the peer, so a `*69` callback waiting on the peer could be missed. Mirroring the hangup teardown closes that gap.

## Test plan

- [x] `go test ./internal/signaling/` (incl. the multi-peer hangup test; the `handleHangup` refactor is behavior-preserving), `-race`, `gofmt`, `go build`.
- [x] Grace tests assert the exact `caller -> callee` end (stronger than the old count-only check); idle/conference tests still assert the `ClearByNumber` path.

### DIFF
--- a/server/internal/signaling/grace_test.go
+++ b/server/internal/signaling/grace_test.go
@@ -19,6 +19,7 @@ func TestNewRelaySetsDefaultGraceWindow(t *testing.T) {
 func TestGraceTimerFiresTeardownAndHangsUpPeer(t *testing.T) {
 	hub := NewHub()
 	tracker := newMockTracker()
+	tracker.peers = map[string]string{"3140001": "3140002"}
 	relay := NewRelay(hub, tracker, nil, nil)
 	relay.GraceWindow = 20 * time.Millisecond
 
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

PR #685, opened 2026-06-02
feat(server): per-line scheduled quiet hours

### LINKED ISSUE #270: feat(server): per-line scheduled quiet hours
**Candidate feature, not necessarily planned.** Filing so it's not lost.

A scheduled window during which a specific line's ringer is automatically silenced. The obvious use case is kids' bedrooms (quiet from bedtime until morning) but adults' rooms work too. Distinct from the household DND master switch (#269), which is an explicit on/off the user operates manually; quiet hours is a time-of-day policy set once and left alone.

Behavior sketch:
- Per-line setting: start time, end time, days of the week.
- During the window: ringer disabled on every handset paired to that line (same ringer-disable semantics as #217).
- Timezone: the household's timezone, since individual handsets don't carry their own timezone today.
- Composes with household DND (#269) and per-device silence (#217): ring only if none of the three is active.

Open questions:
- Any override path? (e.g. a specific caller that can punch through; probably out of scope for v1.)
- UI surface: inline on the Lines list, or a dedicated settings page?
- How is this stored: on the server as line metadata, or on each digitsd?

justinlindh:
## What

Closes #270. Adds a per-line scheduled quiet-hours window: a nightly (or any-time) silent period that a household can set per line, applied on the days they choose.

- New `QuietHours` config on the line `Settings` struct, stored in the existing JSONB settings column. No DB migration. Fields: `Enabled`, `Start`/`End` as "HH:MM" 24-hour strings, and a per-weekday `Days` filter (a fixed `[7]bool` array indexed by `time.Weekday`, which keeps `Settings` comparable so the existing round-trip equality tests still hold).
- Midnight wrap-around: an end time earlier than the start spans midnight (e.g. 22:00 to 07:00). The window is half-open `[Start, End)`, so it releases exactly at the end minute.
- Timezone-aware: the window is evaluated in the owning household's timezone (the timezone already stored on the household).
- Effective silent state: `EffectiveSettingsByNumber` now ORs an active window into `SilentMode`, so the device treats quiet hours exactly like silent mode (calls do not ring; the light still flashes and the call can still be answered). This flows through the existing on-connect settings push in `relay.OnRegistered`.
- Scheduler: a one-minute ticker in signald re-evaluates online lines and pushes updated line settings on window open/close transitions, and only when the effective silent state actually changes, so steady state produces no traffic.
- UI: a quiet-hours editor on the phone detail page (enable toggle, start/end time inputs, day checkboxes), with intercom/dialup and answering-machine theme variants.

## Why

Households want a line to stop ringing overnight without remembering to flip silent mode every evening and morning. Quiet hours automate that on a schedule, per line, in the household's own timezone.

## Out of scope

Caller punch-through/override and per-day distinct times. This is one window applied on the selected days.

## Test plan

- TDD'd the evaluation and `Normalize`/`Merge`: in-window, out-of-window, midnight wrap, day filter, timezone conversion, malformed/empty windows, and JSON round-trip (`internal/line/quiethours_test.go`).
- Scheduler unit tests cover push-on-transition, no-push on steady state, and pruning of offline lines (`internal/signaling/quiethours_test.go`).
- `go test ./...` green; full integration suite green (`-tags integration`, run serially against a local Postgres).
- `golangci-lint run ./...`: 0 issues. `gofmt` clean. `make build` clean.

## Screenshots

Quiet-hours editor, enabled with a Mon-Fri 22:00-07:00 window, rendered in each theme from a local `make dev-up` stack.

### Intercom (default)
intercom

### Dialup
dialup

### Answering machine
answering-machine


justinlindh:
@claude

justinlindh:
@claude

PR #689, opened 2026-06-03
fix(digitsd): flash Pico in setup mode when it does not answer

justinlindh:
## Problem

A fresh device whose RP2040 has never been programmed is silent on first power-up: no setup voice prompts when you lift the hook in WiFi AP/setup mode.

## Root cause

The Pico is auto-flashed **only** by the normal-mode daemon (the POST-FAIL / version-mismatch reflash in `reflashPico`). But a fresh device has no `/data/wifi-configured`, so it boots straight into `--mode=setup` (AP mode), which never runs that path. The result is a chicken-and-egg gap:

- Setup mode opens the UART and PINGs the Pico, gets no response (virgin chip), and the serial subsystem just gives up (`StateFailed`).
- The setup voice loop is gated on `HOOK:OFF` events, which arrive over UART from the Pico. No firmware on the Pico means no hook events, ever.
- So lifting the hook does nothing and the phone stays silent until the user configures WiFi and reboots into normal mode, which finally flashes the Pico.

Reproduced on a V3 board by wiping the RP2040 flash (`openocd ... flash erase_address 0x10000000 0x200000`): `--mode=setup` comes up (web/gpclk0/audio ready) but the serial subsystem PINGs and fails in a loop, with no flash attempt anywhere.

## Fix

Make setup mode flash the Pico when it does not answer, mirroring normal mode:

- `SerialConfig` gains an optional `FlashOnFail func(reason string) error`. When the initial probe fails and a callback is wired, the serial subsystem flashes the firmware once and re-probes before declaring failure. `nil` preserves the old probe-only behavior (recovery mode keeps it).
- Setup mode wires `FlashOnFail` to `runPicoFlashScript`, a shared `flash-pico.sh` runner extracted out of `reflashPico` (same RESCUE retry / FLASHSIZE override / PCB-rev marker, same `SKIP_SERVICE_CONTROL=1` semantics). `flash-pico.sh` already has the RESCUE pre-pass for exactly this virgin-chip case.
- With a flash fallback wired, the pre-flash probe is shortened (a healthy Pico answers on the first attempt, so a long probe just delays the flash a fresh board actually needs); without one the original 10-attempt budget is kept.

The probe/flash/retry orchestration is factored into a pure `attemptSerialBringup` helper so it can be unit-tested without real hardware.

## Testing

- `go test ./...` and `golangci-lint run ./...` clean in `pi/digitsd`.
- New unit tests cover the orchestration: success without flash, flash-then-success, no-flash-configured, flash-error abort, and flash-succeeds-but-still-unreachable.
- Hardware verification on a V3 board with a wiped RP2040 is in progress; results to follow on this PR.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -592,30 +592,42 @@ func readBundledFirmwareVersion() string {
 	return strings.TrimSpace(string(data))
 }
 
-// reflashPico delegates to flash-pico.sh (RESCUE retry, FLASHSIZE override,
-// PCB-rev marker write at 0x101FF000) and re-establishes the serial port.
-// Aborts on serial-reopen failure: nothing else digitsd does works without
-// UART. Returns the reopened port and whether the post-flash PING passed.
-func reflashPico(sp *phone.SerialPort, serialDev string, serialLogger *slog.Logger, reason string) (*phone.SerialPort, bool) {
+// runPicoFlashScript invokes flash-pico.sh (RESCUE retry, FLASHSIZE override,
+// PCB-rev marker write at 0x101FF000) against the bundled firmware.
+// SKIP_SERVICE_CONTROL=1 stops the script from systemctl-stopping us (we ARE
+// digitsd) and from doing its own post-flash PING; callers own the serial port
+// and re-PING themselves. Returns an error if there is no firmware to flash or
+// the script fails, so callers can decide whether to retry the serial link.
+func runPicoFlashScript(reason string) error {
 	if _, err := os.Stat(defaultFirmwarePath); err != nil {
-		slog.Info("reflash: no firmware at path, skipping", "path", defaultFirmwarePath, "reason", reason)
-		return sp, false
+		return fmt.Errorf("no firmware at %s: %w", defaultFirmwarePath, err)
 	}
 	slog.Info("reflash: starting", "path", defaultFirmwarePath, "reason", reason)
-	if err := sp.Close(); err != nil {
-		slog.Warn("reflash: close serial failed", "error", err)
-	}
-	// SKIP_SERVICE_CONTROL=1 stops the script from systemctl-stopping us
-	// (we ARE digitsd) and from doing its own post-flash PING (we hold
-	// the serial port; we'll PING ourselves below).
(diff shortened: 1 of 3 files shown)


justinlindh:
## Hardware verification

Tested on a V3 board with the RP2040 flash erased (`openocd ... flash erase_address 0x10000000 0x200000`), running the fixed binary in `--mode=setup`:

```
subsystem serial: ping failed  x3            (virgin Pico, no PONG)
subsystem serial: Pico unreachable, flashing firmware
reflash: starting  path=/data/digits/firmware.elf  reason=serial-init
=== Digits Pico Flash ===  ...  ** Verified OK **  ...  Rev marker written
reflash: flash script succeeded  reason=serial-init
TX cmd=PING -> RX line=PONG
subsystem serial: Pico reachable after flash   (serial module ready, 16.9s)
subsystem: init complete  ready=4 total=4      (serial was failing before)
TX cmd=STATE:SET:SETUP -> RX line=STATE:SET:OK
```

Setup mode now detects the dead Pico, flashes it, and brings the UART up, so hook events flow and the setup voice prompts can fire on a fresh board. Before this change the serial subsystem reported `StateFailed` and the phone was silent.

justinlindh:
@claude

justinlindh:
Thanks for the review. Addressed the one actionable item and confirmed the other two are intentional.

**3. Trailing `sleep()`:** fixed in 3ce21da2. Both probe loops now skip the sleep after their final attempt, so a fully-failed pre-flash probe no longer burns an extra interval before flashing (and likewise before returning false).

**1. `firstAttempts = 3` reflashing a healthy Pico after factory reset:** keeping as-is. A factory reset leaves the RP2040 programmed, and a healthy already-booted Pico answers the first probe, so it won't miss three. The only way to trip the reflash is a Pico that is both alive and missing 3 consecutive PINGs (~7.5s), and even then the flash is idempotent and verified: the cost is delay, not damage. Raising the budget would slow the genuine virgin-board case (the one this PR exists for) for marginal benefit.

**2. Double `os.Stat`:** intentional, as you noted. `reflashPico`'s early return preserves "do not close the port when there is no firmware," while `runPicoFlashScript`'s own check guards the setup-mode caller that does not pre-stat. Two checks on purpose.

PR #691, opened 2026-06-03
fix(digitsd): auto-launch the captive portal in recovery mode

justinlindh:
## Problem

Connecting an Android phone to recovery mode's `Digits-Recovery` AP does not auto-launch the recovery page, whereas setup/AP mode pops the "sign in to network" page on connect.

## Root cause

Both modes wildcard DNS to `192.168.4.1` and serve the web UI on `:80`, so the probes reach our server in both. The difference is the HTTP response:

- **Setup mode** (`setup.go`) registers the OS captive-portal probe URLs and 302-redirects them to `/`:
  ```go
  for _, path := range []string{"/generate_204", "/hotspot-detect.html", "/connecttest.txt", "/library/test/success.html"} {
      mux.HandleFunc(path, func(w, r) { http.Redirect(w, r, "/", http.StatusFound) })
  }
  ```
  The 302 is what triggers the OS sign-in popup.
- **Recovery mode** (`recovery.go`) never registered those paths, so a probe like `/generate_204` fell through to the `/` static `FileServer` and returned 404. Android reads a 404 as "no portal, no internet" and does not launch the captive UI.

## Fix

Factor the probe list and the redirect into a shared `mountCaptivePortalRedirects(mux, target)` helper (`captive.go`) and mount it in both setup and recovery. Recovery now 302s the probes to its UI root, so the page auto-launches the same way AP mode does. Keeping it in one helper means the two modes can't drift apart again (which is exactly how this gap arose).

## Testing

- `go build ./...`, `go test ./cmd/digitsd/`, and `golangci-lint run ./...` clean in `pi/digitsd`.
- New unit tests assert every probe path 302-redirects to the target and that non-probe paths (e.g. `/api/status`) are left for their own handlers.
- On-hardware check to follow: connect Android to `Digits-Recovery` and confirm the sign-in page auto-opens.

## Related

Part of the same fresh/unpaired-device recoverability work as #689 (setup-mode auto-flash) and #690 (service codes while unpaired).

### DIFF
--- a/pi/digitsd/cmd/digitsd/captive.go
+++ b/pi/digitsd/cmd/digitsd/captive.go
@@ -1,0 +1,27 @@
+package main
+
+import "net/http"
+
+// captivePortalProbePaths are the URLs Android, iOS/macOS, and Windows fetch to
+// decide whether a network is open or behind a captive portal. With AP-mode DNS
+// wildcarded to the device, these requests land on our web server; a 302 to the
+// portal root is what makes the OS pop its "sign in to network" page. Anything
+// else (a 204, or a 404 from the static file server) reads as "no portal", so
+// the page never auto-launches.
+var captivePortalProbePaths = []string{
+	"/generate_204",              // Android, ChromeOS
+	"/hotspot-detect.html",       // iOS, macOS
+	"/connecttest.txt",           // Windows
+	"/library/test/success.html", // older iOS
+}
+
+// mountCaptivePortalRedirects registers a 302 redirect to target for each OS
+// captive-portal probe path. Both setup (AP) mode and recovery mode mount this
+// so the sign-in page auto-launches identically in either.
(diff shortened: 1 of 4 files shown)


justinlindh:
@claude

justinlindh:
Thanks. Addressed note 1; 2 and 3 need no change.

**1. Precedence test:** added in 560cbebd as `TestMountCaptivePortalRedirects_BeatsRootHandler`. It mounts a `/` handler and the probe redirects on the same mux, then asserts `/generate_204` routes to the 302 and the `/` handler is never hit, which is exactly the recovery arrangement that was broken.

**2. Method-agnostic redirect:** leaving as-is. The OS probes are all GET and this is a faithful port of the proven setup-mode behavior; narrowing to GET would be a behavior change for no benefit.

**3. Comment:** no change, as you confirmed it's accurate.

On-hardware Android check on `Digits-Recovery` still to come before merge.

PR #692, opened 2026-06-03
fix(digitsd): re-apply codec mixer state after opening playback

justinlindh:
## Problem

A freshly-flashed V3's earpiece is ~19 dB too quiet. Diagnosed by comparing a quiet unit against a known-good unit with the same onboard TLV320AIC3104 codec (`digitscodec`):

| Control | quiet unit | good unit |
|---|---|---|
| HP DAC | 71 (-23.5 dB) | 100 (-9 dB) |
| HP (analog) | 0 (0 dB) | 5 (+5 dB) |
| HPCOM (analog) | 0 (0 dB) | 5 (+5 dB) |
| PCM (volume) | 81 (level 5) | 81 (level 5) |

Both units **store the identical, correct mixer state** (HP DAC=100, HP=5), and the digitsd volume (PCM) is the same default level. The difference is purely that the quiet unit's earpiece output controls are sitting at codec register defaults rather than the stored values.

## Root cause

The boot-time `alsactl restore` runs as `digitsd.service` `ExecStartPre`, before digitsd opens the playback device. When the render loop subsequently powers up the codec output stage, the TLV320's DAPM-gated controls (`HP DAC`, `HP`, `HPCOM`) come up at register defaults, clobbering the restored values. Re-running `alsactl restore` after the output is powered makes them stick (verified on hardware: a runtime restore moved HP DAC 71 -> 100, HP/HPCOM 0 -> 5).

## Fix

Re-run `alsactl restore` once from the daemon after the render loop has powered the output. Placed immediately before `RestoreVolume()`, so the restore's transient `PCM=100` is overwritten right away by the persisted user volume level. The mixer holds the PCM open continuously for DAC keepalive, so a single re-apply sticks for the process lifetime.

## Testing

- `go build ./...`, `go vet`, `golangci-lint run ./...` clean in `pi/digitsd`.
- The change lives in `main()`'s startup path (no unit-test seam). On-hardware verification: deploy to a V3 and confirm the earpiece comes up loud (`HP DAC=100`, `HP=5`) after a clean reboot with no manual `alsactl restore`. Results to follow on this PR.

## Related

Same fresh-device-recoverability work as #689 / #690 / #691.


justinlindh:
@claude

justinlindh:
## Hardware verification

Deployed to a V3 and rebooted clean (no manual `alsactl restore`). After boot:

```
rootfs:  ro,noatime
HP DAC:  100 [-9.0 dB]    (was 71 [-23.5 dB])
HP:        5 [+5.0 dB]    (was 0 [0 dB])
HPCOM:     5 [+5.0 dB]    (was 0 [0 dB])
PCM:      81 [-23.0 dB]   (volume level 5, preserved)
```

Daemon log confirms the order:
```
INFO mixer re-apply after playback open: applied card=digitscodec
INFO volume restored level=5 max=9 mixer=PCM alsa=81
```

The earpiece comes up at the intended level automatically, and the transient `PCM=100` from the restore is immediately overwritten by `RestoreVolume()` to the persisted level (81). Confirmed the earpiece is audibly louder.

justinlindh:
Thanks, thorough review. Addressed:

**Recovery/setup ordering (the main item):** real gap, fixed in 0ffd21e5. Added the same post-playback-open re-apply to `subsystem/audio.go`. It's gated on a configured `MixerStateFile`, so **recovery** (which sets `/mixer.state`) now gets the DAPM-gated controls applied correctly. **Setup** mode is a no-op here because it configures no mixer state at all (and `digits-setup.service` has no restore, and the embedded state may not be rendered yet on a fresh device) -> that is a deeper, separate gap, filed as #693 with a repro and fix direction. So the onboarding-audibility concern is tracked, not deferred silently.

**Log severity:** good catch, changed to `Warn` in both spots. You're right that `alsactl restore` exits 0 even with the benign UCM warning (the on-hardware reboot logged the success branch), so a non-zero exit is a genuine failure and should be visible.

**PR body "transient PCM=100":** the stored/embedded `v2.state` has `PCM Playback Volume=100`, so `alsactl restore` does write 100 before `RestoreVolume()` sets it to the level-5 value (81). The "100" is accurate; the live 81 in the comparison table is the post-`RestoreVolume` value. Net effect identical either way.

PR #694, opened 2026-06-03
fix(server): extend pairing code TTL and report it to the device

justinlindh:
## Problem

A phone announces a pairing code over voice, says it's valid for several minutes, but the web app rejects it as "invalid or expired."

## Root cause

Two independent constants had drifted apart:

- Server `pairing.CodeTTL = 3 * time.Minute`. The claim query requires `pairing_code_expires_at > NOW()`.
- Device `pairingRefreshInterval = 9 * time.Minute`. Both the spoken countdown (`ceil(9min - elapsed)`) and the reconnect-for-fresh-code timer derived from this.

So a code was actually valid for only the first 3 of every 9 minutes, but the phone kept reading it out as valid for the whole 9. (Tellingly, the device const's own comment already said it "must be shorter than the server-side CodeTTL (10 min)", so the server value, not the device, was the drift.)

Confirmed on hardware (device `0000000053906216`):
```
07:59:06  received pairing code 223864
08:07:24  announced 223864 via voice   (8m18s after receipt)
08:08:20  "pairing code expiring, reconnecting for fresh code" -> 511579
```
The server expired `223864` at ~08:02:06; it was announced as valid more than 5 minutes after it died. Server logs show the periodic "pairing cleanup complete expired_codes:1" nulling it out.

## Fix

Eliminate the drift by making the server's TTL authoritative on the device:

- The server reports the code's validity in the `pairing_code` message (`PairingCodeTTL`, seconds).
- The device drives both its spoken countdown and its refresh timer from that value: it announces the real remaining time and reconnects `pairingRefreshMargin` (60s) before the reported expiry, so the announced code is always still valid when a user types it. Falls back to the fixed interval for servers that do not report a TTL.
- Bump `CodeTTL` from 3 to 10 minutes: a code a user must hear and type needs a comfortable window, and 10m keeps the device's existing ~9m refresh cadence safely inside the validity window.

A TTL (seconds) is sent rather than an absolute timestamp so device/server clock skew cannot distort the countdown.

## Testing

- `go build`, `go test` (unit), `golangci-lint` clean on both `server/` and `pi/digitsd/` (server pairing integration tests run in CI / `make test-integration`).
- Full end-to-end needs the server released to prod (so it sends the TTL) plus the new digitsd on a device. The server-side `CodeTTL` bump alone already makes the existing device cadence correct, and the device change degrades gracefully (fallback) against an old server.

## Rollout note

The device-side precision/refresh change lives in `pi/digitsd` and reaches phones with the next pi image build. The `fix(server)` release restores correct behavior for already-deployed devices on its own, because the 10m TTL matches their existing 9m refresh assumption.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -55,10 +55,16 @@ const iceRestartTimeout = 25 * time.Second
 // checks) within this window, in which case no restart is needed.
 const disconnectDebounce = 4 * time.Second
 
-// pairingRefreshInterval is how often an unpaired device reconnects to
-// obtain a fresh pairing code. Must be shorter than the server-side
-// CodeTTL (10 min) so the code is refreshed before it expires.
+// pairingRefreshInterval is the fallback reconnect cadence for obtaining a
+// fresh pairing code, used only when the server does not report a TTL. When
+// the server sends PairingCodeTTL (it does), the device instead refreshes
+// pairingRefreshMargin before the reported expiry.
 const pairingRefreshInterval = 9 * time.Minute
+
+// pairingRefreshMargin is how far ahead of the server-reported code expiry the
+// device reconnects for a fresh code, so the announced code is still valid
+// while a user is typing it into the web UI.
+const pairingRefreshMargin = 60 * time.Second
 
 // pairingAnnouncementInterval is how long to wait between repeats of the
 // spoken pairing-code sequence while the phone is unpaired and off-hook.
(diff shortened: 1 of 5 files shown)


justinlindh:
@claude

justinlindh:
Thanks. Both observations are non-blocking and I'm leaving them as-is with reasons:

1. **DB-expiry vs device-expiry latency gap:** keeping as-is. The device anchors `expiresAt` to receipt, which trails the server's `now + CodeTTL` write by one WS round trip (sub-second to low-seconds). The 60s `pairingRefreshMargin` is roughly two orders of magnitude larger, so the device always reconnects for a fresh code before the server's `pairing_code_expires_at > NOW()` check fails. The airtight version (have `GenerateCode` return the `expiresAt` it wrote and derive the wire value from that) means changing its signature and the ~17 call sites for precision the margin already swamps. YAGNI.

2. **Stale snapshot on a long (>9m) off-hook listen:** pre-existing and out of scope. The code snapshotted `receivedAt` the same way before this PR, so it's not a regression, and the trigger is holding the handset off-hook through an entire refresh cycle. Not expanding this bug fix to cover it.

PR #696, opened 2026-06-03
fix(digitsd): recovery serial device and the Try Again reboot loop

justinlindh:
## Problems

Two recovery-mode bugs that compound into "recovery is unusable":

### 1. No recovery audio, dead hook switch
Recovery runs as PID 1 (`/sbin/init` -> `/digits-recovery`) with no udev. The `/dev/serial0` that normal boot uses is a udev-created symlink, so it does not exist in recovery; only the kernel device node (`/dev/ttyAMA0` under `disable-bt`) does. But `recoveryRegistrations` opened the default `/dev/serial0`, so the serial subsystem failed, the Pico was unreachable, and `runRecoveryMode` hit its `serial or audio unavailable, voice menu disabled` branch. The phone plays nothing and the hook switch does nothing.

### 2. "Try Again" reboots straight back into recovery
`boot-check.sh` enters recovery whenever `/data/digits/recovery-mode` exists (set by the panic button or a boot-counter trip). Both "Try Again" paths (the web button and voice-menu key 1) cleared only the boot counter and never removed that flag, so the next boot re-entered recovery. The `clearRecoveryFlags` helper that clears both was never wired to either path.

## Fixes

1. `recoverySerialDevice()` opens the raw UART (`/dev/ttyAMA0`, falling back to `/dev/ttyS0`, then the configured `-serial` value), so the serial subsystem comes up and the voice menu runs. Combined with the post-playback-open mixer re-apply already on main (#692), recovery audio plays at the right level.
2. A shared `clearRecoveryBootState()` clears the boot counter **and** removes `/data/digits/recovery-mode`; both try-again paths now call it, so "Try Again" actually exits recovery.

## Testing

- `go build`, `go vet`, `go test ./cmd/digitsd/`, `golangci-lint` clean in `pi/digitsd`.
- This is the recovery-partition binary, so it ships via an image rebuild + reflash of the recovery partition; it can't be verified by deploying to the running rootfs. On-hardware check after reflash: hold `*` at boot to enter recovery, confirm the menu speaks on hook-lift, and confirm "Try Again" boots normally instead of looping.

## Escaping the loop on an already-affected device
Until this ships, the web "Try Again" loops. Escape via the recovery web UI Factory Reset (wipes `/data`, clearing the flag), or by mounting the `data` partition offline and deleting `digits/recovery-mode`.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -881,10 +881,24 @@ func playPairingAnnouncement(mixer *audio.Mixer, code string, expiresAt time.Tim
 	mixer.PlayOnce(unitClip)
 }
 
+// recoverySerialDevice returns the raw UART node for the Pico. Recovery mode
+// runs as PID 1 without udev, so the /dev/serial0 symlink that normal boot
+// relies on does not exist; only the kernel-created device node does. Under
+// disable-bt the Pico is on the PL011 (/dev/ttyAMA0). Falls back to the
+// configured -serial value if neither raw node is present.
+func recoverySerialDevice() string {
+	for _, dev := range []string{"/dev/ttyAMA0", "/dev/ttyS0"} {
+		if _, err := os.Stat(dev); err == nil {
+			return dev
+		}
+	}
+	return *serialDev
+}
+
 func recoveryRegistrations() ([]subsystem.Registration, *subsystem.WebModule, *subsystem.SerialModule, *subsystem.AudioModule) {
 	web := subsystem.NewWebModule()
 	gpclk0 := subsystem.NewGPCLK0Module()
-	serial := subsystem.NewSerialModule(subsystem.SerialConfig{Device: *serialDev, Baud: 115200})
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

justinlindh:
Thanks, and good catch on the stacked-mount footgun being the deeper cause.

**Documented it** in d39638ad: added a comment on `clearRecoveryFlags` spelling out that its mount/remove is a no-op when /data is already mounted (the PID-1 recovery path), so callers there must use `clearRecoveryBootState` first. Left the function otherwise as-is since it's now off the critical path for the loop, and `doReboot`'s call to it is harmless belt-and-suspenders.

**On the other notes (all non-blocking, leaving as-is):**
- The `os.Stat` UART probe isn't hardware detection, agreed. It's correct only because the image always sets `disable-bt` (PL011 = ttyAMA0 = Pico). If a future variant drops `disable-bt` this would pick wrong silently, but that's a fixed invariant of the image today.
- The final `*serialDev` fallback (default `/dev/serial0`) can't recover in recovery since that symlink is absent without udev. It's a deliberate belt-and-suspenders for an explicit `-serial` raw node; the implicit default failing is the same as today's behavior.

PR #697, opened 2026-06-03
feat(server): collapse quiet-hours config until the toggle is on

justinlindh:
## What

On the per-line drill-down page, the quiet-hours window (From/To), day pickers, and Save button were always rendered, just dimmed to 55% opacity when the feature was off. The controls stayed fully interactive even when disabled, which read as clutter and let you edit a window that had no effect.

Now the config collapses until the **Scheduled quiet hours** toggle is checked:

- Toggle OFF: From/To/Days hidden.
- Toggle ON: From/To/Days shown.
- Reactive: toggling shows/hides instantly, no reload.
- Initial render reflects the saved state.
- The enable toggle and Save button stay visible in both states, so you can still turn quiet hours off and persist it.

## How

A CSS `:has()` rule on `#quiet-hours-section` drives the show/hide off the live checkbox state. No JS. Because the rule keys on the rendered `checked` attribute, the first paint already matches the saved setting, and because it keys on the live checkbox, toggling reacts instantly. The htmx outerHTML swap on save re-renders the section with the correct `checked`, so the state survives a save.

Only the From/To/Days inputs collapse. The enable checkbox and the Save button live in the `<form>` but outside the collapsing wrapper, so unchecking and saving the disabled state always works. The fields use `display:none` rather than `disabled`, so their values still submit and the handler keeps reading the full window on save.

Applied to all three themes:

- intercom: `server/internal/web/static/digits.css`
- dialup: `server/internal/web/static/dialup.css`
- answering-machine: `server/internal/web/static/answering-machine.css`

The answering-machine markup reuses the `.am-voicemail-fields` class, so that rule is scoped to `#quiet-hours-section` and never touches the voicemail section's identically-classed wrapper.

## Verification

Built and ran locally (`make dev-up`), drove with agent-browser. For each theme, confirmed: From/To/Days hidden when off, appear when toggled on, initial state matches saved setting, the disable path (uncheck a genuinely-enabled line, click Save) persists `enabled: false` to the DB, and the voicemail section's fields stay visible. `make test` and `golangci-lint run ./...` both clean. No Go changed (template + CSS only).

Screenshots in the PR comments: collapsed vs expanded per theme, and the corrected collapsed state showing Save remains reachable.

### DIFF
--- a/server/internal/web/static/answering-machine.css
+++ b/server/internal/web/static/answering-machine.css
@@ -2483,3 +2483,12 @@ body.am .pair-mode-field legend {
   align-self: flex-start;
   margin-top: 4px;
 }
+
+/* Quiet hours: collapse the window/day config until the enable toggle is on.
+   Scoped to the quiet-hours section so it never touches the voicemail
+   section's identically-classed field wrapper. Driven off the rendered
+   checkbox state, so the initial paint matches the saved setting and
+   toggling reacts instantly with no JS. */
+#quiet-hours-section form:not(:has(input[name="enabled"]:checked)) .am-voicemail-fields {
+  display: none;
+}
--- a/server/internal/web/static/dialup.css
+++ b/server/internal/web/static/dialup.css
@@ -751,6 +751,13 @@ body.dialup:not(.crt-all) .dialup-window {
   inset: 2px;
   background: var(--dialup-accent-red);
 }
+
+/* Quiet hours: collapse the window/day config until the enable toggle is on.
+   Driven off the rendered checkbox state, so the initial paint matches the
+   saved setting and toggling reacts instantly with no JS. */
+#quiet-hours-section form:not(:has(input[name="enabled"]:checked)) .quiet-hours-fields {
+  display: none;
+}
 .radio-card__desc {
   margin: 3px 0 0 18px;
   font-size: 11px;
--- a/server/internal/web/static/digits.css
+++ b/server/internal/web/static/digits.css
@@ -807,6 +807,13 @@ hr { border: 0; border-top: 1px solid var(--rule); margin: 0; }
   background: var(--brass);
   border-radius: 50%;
 }
+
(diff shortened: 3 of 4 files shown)


justinlindh:
### Before / after, all three themes

The "before" is the collapsed (toggle off) state, which is the new behavior. Previously the same fields were always rendered at 55% opacity regardless of the toggle.

**intercom**, off (collapsed) then on (expanded):

intercom collapsed
intercom expanded

**dialup**, off then on:

dialup collapsed
dialup expanded

**answering-machine**, off then on. Note the adjacent Answering Machine section keeps its own fields visible, confirming the scoped selector:

am collapsed
am expanded

justinlindh:
@claude

justinlindh:
Fixed the blocking issue in 2acdbb30. You were right: the Save button sat inside the collapsing wrapper, so unchecking the toggle on an enabled line hid Save and trapped the line in the on state.

Fix: moved the Save button (and the enable toggle was already outside) out of `.quiet-hours-fields` / `.am-voicemail-fields` in both `quiet-hours-section` and `am-quiet-hours-section`, so only From/To/Days collapse. The CSS selectors did not need to change since they still key on those wrappers.

Re-verified the disable path on all three themes against a line that was genuinely enabled (valid 22:00 to 07:00 window saved through the UI first, since `Normalize()` drops an enabled flag with no times): uncheck leaves Save reachable, clicking Save persists `enabled: false` to the DB. Screenshots of the collapsed state now showing Save:

intercom collapsed with save
dialup collapsed with save
answering-machine collapsed with save
answering-machine expanded

On the two minor notes: `:has()` browser support is fine for the evergreen targets here, and as you noted the degraded fallback (fields always visible) is the safe one. On toggle UX drift with voicemail's instant-persist endpoint: skipping that as out of scope for this change. Quiet hours is an explicit-Save form by design and adding an auto-persist endpoint is a separate behavior change, not cleanup this PR should smuggle in.

PR #698, opened 2026-06-03
fix(image): make recovery mode audio, auto-reset, and AP-independent menu work

justinlindh:
Recovery mode shipped a set of latent failures that left both the handset voice menu and the advertised auto-factory-reset non-functional. A code audit found them with file:line; each was re-confirmed against current code before changing. These are image-build plus recovery-code changes that ship via a fresh image build and reflash.

### F3 (highest priority): recovery audio kernel modules never loaded
`tools/build-image.sh` decompresses every recovery module to a FLAT layout (`/lib/modules/$KVER/<basename>.ko`), and the WiFi chain in `kernmods.go` loads by basename to match. The audio chain, however, loaded by nested kernel-tree paths (`kernel/sound/core/snd.ko`, ...), so every audio `insmod` resolved to a nonexistent file and silently failed. The `digitscodec` card never registered, `NewPlayback` failed, and the mixer was nil.

- `kernmods.go`: load the audio chain by basename to match the flat layout, and reorder leaf-deps-first (insmod does no dependency resolution).
- Added the `i2c-bcm2835` bus controller to the chain, loaded before the codec I2C driver (`snd-soc-tlv320aic3x-i2c`) so the TLV320AIC3104 control bus can bind. Without it the codec never probes.
- `build-image.sh`: added `i2c-bcm2835.ko.xz` to the audio module copy list and an `i2c-bcm2835.ko:` entry to `modules.dep`.

Load order is leaf-deps-first: snd, snd-timer, snd-pcm, snd-pcm-dmaengine, snd-compress, snd-soc-core, regmap-i2c, i2c-bcm2835, snd-soc-bcm2835-i2s, snd-soc-tlv320aic3x, snd-soc-tlv320aic3x-i2c, snd-soc-simple-card-utils, snd-soc-simple-card.

### F8: alsactl and /mixer.state missing on the recovery partition
`recoveryRegistrations` sets `MixerStateFile:"/mixer.state"` and the audio subsystem runs `alsactl restore -f /mixer.state`, then re-applies per-card after the output powers on (it probes the card name with `amixer`). Neither `alsactl` nor `amixer` was copied to recovery `/bin`, and `/mixer.state` was never staged.

- `build-image.sh`: added `alsactl` and `amixer` to the recovery tool copy loop (libs and the `/usr/share/alsa` tree are already handled), and stage the per-PCB mixer state at `/mixer.state` using the same canonical `$MIXER_STATE` selected for `/data` in step 14b. Recovery keeps its own copy because `/data` is reformatted during factory reset.

### F4: WiFi-AP was Required, halting recovery before the WiFi-independent voice menu
`recoveryRegistrations` registered `WiFiAPModule` with `Required: true`; a failed AP errored `mgr.Run`, which short-circuits to `syncAndHalt()` before `runRecoveryMode` ever starts the voice loop. The handset menu needs only serial plus audio.

- `main.go`: drop `Required` on the recovery AP so a failed AP warns and continues. Nothing depends on `wifi-ap`, so the layering is unaffected.
- `build-image.sh`: ship `rfkill` (used by `unblockWifi`) both as the `rfkill` package in the rootfs and through the recovery tool copy loop.

### F5: app/service-code factory reset was never honored in recovery
`triggerFactoryReset` writes `bootcount.AutoFactoryResetFlag` and sets the boot counter to its threshold so recovery should auto-run the reset, but nothing in `recovery.go` read the flag, so the user landed at the manual menu.

- `recovery.go`: on recovery startup, if the flag is present, clear it first (so a failed/interrupted reset falls back to the manual menu next boot instead of looping) and run `doRecoveryFactoryReset` directly. Works even when audio or serial failed, since the reset path guards both.

### F7 (low): error_tone played but error_tone.wav did not exist
`doRecoveryFactoryReset` plays `error_tone` on failure, but there was no such tone, so failures were silent.

- Added `pi/tones/error_tone.wav`: 44100 Hz mono S16_LE descending double-beep (440 then 311 Hz), matching the format of the other recovery tones. The build already copies every `pi/tones/*.wav` to the recovery partition.

### PATH determination
Recovery runs as PID 1 after `switch_root` and inherits the environment from the initramfs `/init`. Its `PATH` is not contractually guaranteed to include the `/bin` and `/sbin` where the recovery partition stages its tools, and every `exec.Command` in the recovery path (`insmod`, `ip`, `alsactl`, `amixer`, `mkfs.ext4`, `hostapd`, `dnsmasq`, `rfkill`, ...) is invoked by bare name. Rather than rely on whatever the initramfs left behind, `main.go` now pins `PATH=/sbin:/bin:/usr/sbin:/usr/bin` and `LD_LIBRARY_PATH=/lib` at the top of the recovery PID-1 branch.

## Not in this PR
- F6 (rootfs-snapshot vs recovery-partition headroom) is a build-time fragility, not a runtime failure, and was explicitly out of scope.

## Verification
- `cd pi/digitsd && make build` cross-compile to linux/arm64: passes.
- `go test ./...`: passes. `go vet ./...`: clean. `golangci-lint run ./...`: 0 issues.
- `bash -n tools/build-image.sh`: valid (shellcheck is not installed on this host; the shell edits were reasoned through manually).
- Recovery is a PID-1 / switch_root / on-hardware path. These changes are verified on-device only after an image rebuild and reflash; that has not happened yet.


justinlindh:
@claude please review

justinlindh:
Thanks for the review. Walked the three on-device items and the two minor notes; no code changes needed, but I verified the one that could have been a silent failure.

**Item 1 (is `i2c-bcm2835` actually a module?): verified on a live phone, it is a module.** This was the right thing to check first. On `6.6.51+rpt-rpi-v8`:

```
$ zgrep CONFIG_I2C_BCM2835 /boot/config-6.6.51+rpt-rpi-v8
CONFIG_I2C_BCM2835=m
$ find /lib/modules/$KVER -name 'i2c-bcm2835*'
/lib/modules/6.6.51+rpt-rpi-v8/kernel/drivers/i2c/busses/i2c-bcm2835.ko.xz
$ lsmod | grep i2c_bcm2835
i2c_bcm2835            16384  0
```

So `=m`, the `.ko.xz` exists at the path the copy loop reads, and it is loaded at runtime in normal boot. The copy-list entry, the `modules.dep` entry, and the kernmods load are all load-bearing, and the comment's causal claim (codec I2C bus never binds without it) holds. No silent-skip path here.

**Item 2 (error_tone.wav format):** byte-verified against the other recovery tones. Identical `fmt ` chunk: PCM (0x0001), 1 channel, 44100 Hz, 16-bit, matching e.g. `reset_complete.wav`. It even has a cleaner header (data chunk immediately after fmt, no LIST/INFO metadata). The mixer's `loadWAV` walks chunks to find `data`, so either layout parses fine. Still a "listen once on-device" item like the other tones, but no rate/pitch mismatch relative to the existing recovery tones.

**Item 3 (alsactl config deps):** the recovery section already copies the full `/usr/share/alsa` tree and `/etc/asound.conf` (build-image.sh, the alsa block just above the tool loop). That is the same config the embedded `digits_playback`/`digits_capture` virtual PCMs already rely on, so `alsactl restore -f /mixer.state` has its config tree. Sanity check stands for the reflash.

**Minor (dead-weight `modules.dep` entry, redundant `LD_LIBRARY_PATH`):** keeping both, for the reasons you noted: the `modules.dep` line keeps the file consistent for the modprobe fallback path, and the per-call `LD_LIBRARY_PATH=/lib` in `insmod` is defensible since `insmod` can be reached outside the PID-1 branch. Agreed neither is what makes the load work.

The on-hardware reflash verification remains the only real outstanding risk, as called out in the PR body.

justinlindh:
Thanks, thorough review. No code changes: you confirmed the logic and load order, and the three "confirm on device" items are exactly the reflash-verification caveats the PR calls out, not code defects.

On them:
1. **Is `i2c-bcm2835` actually a module?** Can't check right now (the test device is mid-reflash and off the network), so this becomes a `dmesg` check after reflash. Either outcome leaves the PR correct: if it's built-in (`=y`), the build's copy hits the warn-and-skip branch and is a harmless no-op, with the basename fix being the operative change; if it's `=m`, the copy is necessary. The basename fix alone already makes the staged modules loadable, so audio is not solely dependent on the i2c addition.
2. **`error_tone.wav` format:** listen-once item at reflash; it was authored to match the other recovery tones (44100Hz mono S16_LE).
3. **`alsactl` config deps:** the `/usr/share/alsa` tree is copied to the recovery partition, and `-f /mixer.state` uses the explicit state file; will sanity-check for a missing-`alsa.conf` warning at reflash.

Minor notes both left as-is with your reasoning: the `i2c-bcm2835` `modules.dep` entry is kept for modprobe-fallback consistency with the rest of the file (insmod uses explicit paths, so it's inert on the current path), and the `LD_LIBRARY_PATH` redundancy is harmless and defensible since `insmod` can be reached outside the PID-1 branch.

Verification remains on-hardware after an image rebuild + reflash, as noted in the PR.

justinlindh:
Added commit `ede340cc` for the real reason "Try Again" loops, which the earlier fixes (and #696) missed.

`runRecoveryMode` sets the Pico's persisted phase byte to `RECOVERY` (`recovery.go:176`, for the LED). The normal-boot panic-button handler reads that byte and, if it's `RECOVERY`, re-writes `/data/digits/recovery-mode` and reboots back into recovery. Both try-again exits cleared the `/data` flag + boot counter but never cleared the Pico phase, so: try-again -> reboot -> normal boot reads `PHASE_RECOVERY` -> panic handler recreates the flag -> recovery. Loop. Factory reset already dodged this via `StateSet("SETUP")` before its reboot (which is why factory reset escapes cleanly); the try-again paths did not.

It only surfaced once recovery serial actually worked: with the prior broken `/dev/serial0`, `StateSet("RECOVERY")` never ran (nil `sp`), so the phase was never set and try-again appeared fine. The serial fix in this PR made recovery set the phase, exposing the missing clear-on-exit.

Fix: `clearRecoveryPhase(sp)` (StateSet to a non-recovery value) on both try-again exits; normal boot re-sets the correct paired/unpaired phase from the token. Build/vet/test/lint clean. Still verified on-hardware only after reflash.

justinlindh:
@claude

PR #705, opened 2026-06-04
fix(digitsd): honor the panic button (* at boot) from setup/AP mode

justinlindh:
## Problem

Holding `*` at power-on should enter recovery. On a device with no WiFi configured, the LED flashes as if entering recovery, but ~25s later it proceeds to AP config mode instead. The user is not holding `*` after the flash (and shouldn't have to).

## Root cause

The panic-button recovery entry (query the Pico's persisted phase byte, and if it's `PHASE_RECOVERY` flag recovery for the next boot and reboot) exists **only in the normal-mode startup path** (`main.go`, after the setup branch returns). A device with no `/data/wifi-configured` boots into setup/AP mode (`--mode=setup`), which runs `runSetupMode` and returns without ever querying the phase. So:

- Firmware sees `*` held at its boot scan, writes `PHASE_RECOVERY`, fast-blinks the LED (this is the flash you see).
- The Pi boots into setup mode, ignores the phase, brings up the AP, and `StateSet("SETUP")` flips the LED back (the ~25s mark).

`boot-check.sh` doesn't read the Pico phase either (it keys off the `/data` flag and boot counter), so nothing in the AP-mode boot acts on the panic.

## Fix

Extract the phase query and the recovery-entry into `queryPicoPhase` and `enterPanicRecovery`, and run the same check in setup mode after the serial subsystem has brought the Pico up. Setup mode is pre-pairing, so it passes an empty device token (`StateSet("UNPAIRED")`). Normal mode is refactored to use the shared helpers (no behavior change).

After this, holding `*` at power-on enters recovery whether or not WiFi is configured, with no need to keep holding the button.

## Testing

- `go build`, `go test ./cmd/digitsd/`, `go vet`, `golangci-lint` clean in `pi/digitsd`.
- This is the main rootfs binary (setup mode runs from `/usr/local/bin/digitsd` via `digits-setup.service`), so it ships via the normal pi release / OTA, NOT a recovery-partition reflash. On-hardware check: on a no-WiFi device, hold `*` at power-on and confirm it enters recovery (and stays) instead of falling through to AP config.

## Relationship to the recovery work
Independent of the recovery-partition fixes (#698 merged, #703 open). This one is about *entering* recovery from setup mode; those are about recovery *functioning* and *exiting* cleanly. Requires a flashed Pico (the firmware must be present to scan the keypad and write the phase).

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -1085,7 +1085,9 @@ func main() {
 		// Setup mode is pre-pairing, so pass an empty device token.
 		if subsystem.IsReady(serial) {
 			if sp := serial.Port(); sp != nil {
-				if phase, err := queryPicoPhase(sp); err == nil && phase == phone.PhaseRecovery {
+				if phase, err := queryPicoPhase(sp); err != nil {
+					slog.Error("setup: phase query failed, proceeding without panic-button check", "error", err)
+				} else if phase == phone.PhaseRecovery {
 					enterPanicRecovery(sp, "")
 				}
 			}
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -975,6 +975,43 @@ func setupRegistrations() ([]subsystem.Registration, *subsystem.WebModule, *subs
 	return regs, web, serial, audio
 }
 
+// queryPicoPhase reads the Pico's persisted phase byte, retrying briefly since
+// the link may still be settling right after POST.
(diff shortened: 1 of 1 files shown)


justinlindh:
@claude

justinlindh:
Addressed the log-symmetry nit in af3636b4: the setup-mode phase check now logs `slog.Error("setup: phase query failed, proceeding without panic-button check")` on a query error, matching the normal-mode breadcrumb, so a RECOVERY-flashed device that falls through to AP config because the query failed is diagnosable from logs. The two notes (redundant SetThreshold, no explicit sync before reboot) are pre-existing/shared-path behavior, left as-is.

PR #706, opened 2026-06-04
fix(digitsd): cap spoken pairing-expiry minutes at 9 so the number is audible

justinlindh:
## Problem

Pairing a new device reads the code, then says "this code expires in **[silence]** minutes" instead of "expires in **X** minutes". Regression.

## Root cause

`playPairingAnnouncement` clamps `minutesLeft` to a max of 10 and plays `spoken_<minutesLeft>`, but the spoken number clips only exist for `spoken_0`..`spoken_9` (no `spoken_10`). When the pairing `CodeTTL` was raised from 3 to 10 minutes (#694), a freshly issued code computes `ceil(time.Until(expiresAt)) == 10` for its first minute, so it plays the nonexistent `spoken_10` and the number is silent. Before #694 the count was based on the 9-minute refresh interval and never reached 10, so this never surfaced.

## Fix

Cap `minutesLeft` at 9 (the highest available clip). A fresh 10-minute code reads "9 minutes" for its first minute, which is true and conservative (the code is valid for at least that long); after the first minute it tracks accurately (8, 7, ...). Robust to future TTL changes since it can never exceed the available clips.

## Testing

- `go build`, `go test ./cmd/digitsd/`, `golangci-lint` clean in `pi/digitsd`.
- On-hardware: pair a device and confirm the spoken count is present ("expires in 9 minutes") rather than silent.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -867,10 +867,14 @@ func playPairingAnnouncement(mixer *audio.Mixer, code string, expiresAt time.Tim
 		mixer.PlayOnce("spoken_" + string(ch))
 	}
 	minutesLeft := int(math.Ceil(time.Until(expiresAt).Minutes()))
+	// Spoken number clips only exist for 0-9, so cap at 9. A fresh code (TTL
+	// 10m) reads "9 minutes" for its first minute, which is true and
+	// conservative (it is valid for at least that long); without the cap it
+	// would play the nonexistent spoken_10 and the number would be silent.
 	if minutesLeft < 1 {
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

justinlindh:
Took the suggestion in 8a6692a6: extracted `pairingAnnouncementClips` as a pure function (mirroring the `voicemail.go` sequence builders) and added a test pinning the boundary at 9 (10+ caps to `spoken_9`, never `spoken_10`; 1 uses the singular unit; 0 floors to 1; code digits expand in order). So a future TTL bump can't silently reintroduce the silent number. Build/test/lint clean.

PR #725, opened 2026-06-09
fix(firmware): off-hook timeout, keypad debounce, and breathing LED flash

justinlindh:
## What

Fixes three verified code-review findings in the Pico H firmware (`firmware/src/`).

**Finding 1 (MEDIUM): DIALING off-hook timeout never fires for star/pound-only input.** `firmware/src/phone_fsm.c:526` (with `412-418`). Off-hook enters DIAL_TONE; the first keypress transitions to DIALING regardless of key. The DIALING timeout only fired when `s_digits_len > 0`, and `s_digits_len` increments only for `'0'-'9'`. A first key of `*` or `#` (a service-code or recovery prefix) leaves `s_digits_len == 0`, so neither the DIALING timeout nor the now-exited DIAL_TONE timeout applied, and the phone sat silently in DIALING until the handset was cradled. Fix: drop the `s_digits_len > 0` guard so the timeout fires on time since DIALING entry (`s_dialing_start_ms`, set on entry, not reset per digit) regardless of digit count, transitioning to BUSY exactly like the existing path. Sequences containing at least one numeric digit already worked and are unchanged; the completed-dial path still short-circuits via `!s_dial_sent`. Reuses the existing `TIMEOUT:DIAL_TONE` UART event so digitsd applies its off-hook permanent-signal treatment instead of seeing a new unhandled event.

**Finding 2 (MEDIUM): fast distinct-digit dialing drops keys.** `firmware/src/keypad.c:53-68`. The debounce timer was updated only on accept and never reset on release, so a distinct key B pressed within `KEYPAD_DEBOUNCE_MS` (80ms) of key A's accept was dropped because the window was still measured from A's accept time. Fix: track the last debounced state change and confirm releases. A genuinely new key after a confirmed release is accepted immediately (it cannot be bounce of a prior key because a release intervened), while same-key bounce is still rejected: a brief mid-press open does not persist long enough to confirm the release, so a bounce re-close remains a held no-op. Holding a key still produces no auto-repeat. State walk `no key -> A -> no key -> B` now accepts both; `A -> brief open -> A` (bounce) still yields a single A.

**Finding 3 (LOW): one-tick full-bright flash on breathing entry.** `firmware/src/led.c:99-103` (with `176-186`). `led_set_mode(LED_MODE_BREATHING)` fell into the generic else branch that forces `PWM_WRAP` (full brightness); `led_update` then started the breathe LUT at phase 0 (dim), producing a visible full-bright flash on every breathing entry, including at boot (`main.c` PHASE_PAIRED). Fix: on breathing entry, seed the output to `s_breathe_lut[0]` (the ramp's dim start) so the entry level matches the first `led_update` step and there is no flash.

**Finding 4 (INVESTIGATE): V1 L298N ENABLE handling. Decision: skip, no code change.** The concern was that `set_hbridge`/`stop_hbridge` assume IN1=IN2=LOW silences the coil, which on the V1 L298N depends on the ENABLE pin that the firmware never drives. Investigation of the V1 hardware (`hardware/pcb/v1/ERRATA.md`, `hardware/pcb/v1/PLAN.md`, `docs/build/wiring.md`) shows the firmware cannot and need not drive ENABLE on V1:

- V1's ringer is an external L298N module on header J5, which breaks out only IN1, IN2, +12V, and GND (`PLAN.md` J5 row; `ERRATA.md` root-cause entry). There is no ENABLE conductor routed to the Pico, so there is no pin for the firmware to control.
- On the standard L298N module the ENA/ENB enable inputs carry onboard jumpers tying them to +5V, so the output stage is permanently enabled. With ENABLE held high, IN1=IN2=LOW puts the bridge in brake-to-ground (both low-side devices on), which clamps the transformer primary at 0V: no AC drive, silent. The firmware assumption holds.
- The Pico GPIO pins on the V1 profile (`firmware/src/board.c`: `ringer_in1_pin = 11`, `ringer_in2_pin = 15`) match the documented J5 IN1/IN2 wiring, confirming the two control lines are the only ringer signals under firmware control.

No safe, verifiable code change is warranted: there is no V1 ENABLE wire to drive, and the existing both-LOW silence behavior is correct for both the V1 L298N (brake while enabled) and the V2 DRV8871 (coast). This is a concrete hardware reason, not a deferral.

## Why

These are correctness bugs in production phone hardware: a phone that hangs in DIALING after a `*`/`#` prefix, dropped digits during normal fast dialing, and a cosmetic LED flash on every breathing entry.

## Test plan

- `make firmware` (Docker build) succeeds cleanly, ELF and UF2 produced.
- Self-verified the FSM, keypad, and LED state machines by tracing the affected transitions, since hardware flashing is not available in this environment:
  - DIALING timeout: zero-digit (`*`/`#`-only) now transitions to BUSY at `DIAL_TIMEOUT_MS`; numeric partial dials unchanged; completed 7-digit dial still short-circuits via `!s_dial_sent`.
  - Keypad: `no key -> A -> no key -> B` accepts both; same-key bounce yields one event; held key gives no auto-repeat; double-tap of the same digit after a confirmed release registers twice.
  - LED: breathing entry now writes `s_breathe_lut[0]` (dim), matching the first `led_update` step, eliminating the full-bright tick.

### DIFF
--- a/firmware/src/keypad.c
+++ b/firmware/src/keypad.c
@@ -10,8 +10,14 @@
 #define KEYPAD_SETTLE_US 5
 #define KEYPAD_DEBOUNCE_MS 80
 
+// s_last_key is the key currently believed to be held (the last one accepted),
+// or '\0' once a release has been confirmed. s_change_time marks the last
+// debounced state change (accept or confirmed release); a new transition is
+// honored only after KEYPAD_DEBOUNCE_MS has elapsed since it, which rejects
+// contact bounce on a single key without measuring distinct presses against an
+// unrelated earlier key's accept time.
 static char s_last_key = '\0';
-static absolute_time_t s_last_press_time;
+static absolute_time_t s_change_time;
 
 void keypad_init(void) {
     for (int i = 0; i < KEYPAD_NUM_ROWS; ++i) {
@@ -27,7 +33,7 @@ void keypad_init(void) {
     }
 
     s_last_key = '\0';
-    s_last_press_time = get_absolute_time();
+    s_change_time = get_absolute_time();
 }
 
 char keypad_scan_raw(void) {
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude

PR #726, opened 2026-06-09
fix(server): harden client-IP resolution and Google OAuth, plus latent concurrency fixes

justinlindh:
## What

Five verified code-review fixes in the signaling server: two security hardening fixes and three latent concurrency fixes.

**1. Client IP resolution (security).** `httputil.ClientIP` trusted `X-Forwarded-For` whenever `RemoteAddr` was private and returned the leftmost entry. Behind Traefik in the cluster, `RemoteAddr` is always the private proxy IP, so any client could send its own `X-Forwarded-For` and control the rate-limit key, defeating every limiter (magic-link, Google login, pairing, invite, WebSocket upgrade). `ClientIP` now takes a trusted-proxy hop count and walks that many entries from the right, since each trusted proxy appends the address it received the connection from. The rightmost-minus-hops entry is the only one the chain vouches for. Default is one hop (a single Traefik), overridable via `SIGNALD_TRUSTED_PROXIES`; zero ignores `X-Forwarded-For` entirely.

**2. Google OAuth email verification (security).** The OAuth callback never checked `verified_email`, then matched an existing user by email and linked the Google ID. An attacker could set a victim's address as an unverified email on their own Google account and take over the victim's magic-link account. The callback now parses `verified_email` and rejects the login with 403 when false, before any user lookup, link, or session creation.

**3. Conference snapshot (data race).** `ConferenceByPhone` returned the live `*Conference` pointer, which callers read after the tracker mutex was released while other methods mutate the same struct under lock. It now returns a deep copy, factored into a `snapshotLocked` helper shared with the existing `Snapshot`. All callers only read fields.

**4. Hub reconnect channel close (message loss).** On a same-hardware-ID reconnect, `Hub.Register` drained `old.Send` before closing it, discarding any queued outbound frame. The double-close the drain guarded against cannot happen: replacing the slot in place makes the old conn invisible to any later `Unregister`, and every send happens under `h.mu`. The channel is now closed unconditionally, so the write pump exits on its `!ok` branch after flushing what it has buffered.

**5. Redis subscriber goroutine leak.** The subscriber goroutine selected on `ctx.Done()` for the receive from Redis but used an unguarded blocking send to the consumer channel. A stalled consumer with a full buffer would leak the goroutine on shutdown. The send is now wrapped in a `select` with a `ctx.Done()` case.

## Why

Fixes 1 and 2 are exploitable account-takeover and rate-limit-bypass vectors in production. Fixes 3 through 5 are latent concurrency bugs (data race, silent message loss on reconnect, goroutine leak on shutdown).

## Test plan

- `make test`: pass
- `make test-integration`: pass (with `-race`, covering the conference snapshot and hub close fixes)
- `golangci-lint run ./...`: clean (no new issues)
- `make build`: pass

New and updated unit tests cover direct connections, single and multi trusted-proxy XFF resolution, spoofed leftmost entries being ignored, zero trusted proxies, the unverified-email rejection path, and same-hardware reconnect preserving a queued frame while closing the old channel.

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -255,6 +255,7 @@ func run(ctx context.Context) error {
 		AdminSecret:       cfg.AdminSecret,
 		DevMode:           cfg.DevMode,
 		WSRateLimitPerMin: cfg.WSRateLimitPerMin,
+		TrustedProxies:    cfg.TrustedProxies,
 	})
 	if err != nil {
 		return fmt.Errorf("create handler: %w", err)
--- a/server/internal/auth/google.go
+++ b/server/internal/auth/google.go
@@ -106,12 +106,13 @@ func (g *GoogleAuth) HandleCallback(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var info struct {
-		ID    string `json:"id"`
-		Email string `json:"email"`
-		Name  string `json:"name"`
-	}
-	if err := json.Unmarshal(body, &info); err != nil {
+	info, err := parseGoogleUserinfo(body)
+	switch {
+	case errors.Is(err, errUnverifiedEmail):
+		slog.WarnContext(r.Context(), "auth: google callback rejected unverified email", "google_id", info.ID)
+		http.Error(w, "Google account email is not verified", http.StatusForbidden)
+		return
+	case err != nil:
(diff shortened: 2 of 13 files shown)


justinlindh:
@claude

PR #727, opened 2026-06-09
fix(digitsd): reconnect off the main loop and resilience hardening

justinlindh:
## What

Fixes six verified code-review findings in the Pi daemon. The dominant fix moves WebSocket reconnection off the main select loop so it no longer blocks event servicing and shutdown. The rest harden the updater, recovery mount handling, debug capture, and the audio filter path, and close a data race on the signaling client pointer.

## Why

Per finding:

- **Reconnect blocked the main loop (HIGH).** `cmd/digitsd/main.go` ran an inner backoff loop with `time.Sleep` (exp backoff capped 60s) directly inside the main `select`. While sleeping the loop could not service `sp.Events()` (UART hook/key events) or `quit`, so events queued or dropped and `systemctl stop` blocked up to 60s. Reconnection now runs in `reconnectLoop` on its own goroutine; it does the backoff, connect, and post-reconnect resume/teardown, then hands the live client back over a buffered channel. A `reconnecting` flag nils the dead client's already-closed `Done()` out of the select so the loop neither busy-spins nor blocks and stays free to service events and shutdown. The active-call fast backoff vs idle exponential backoff (`nextReconnectBackoff`/`IsCallActive`), watchdog petting (already a separate goroutine), and clean shutdown are preserved.

- **Updater HTTP could hang forever (HIGH).** `internal/updater/updater.go` shared an `http.Client{}` with no timeout and `Download` used `client.Get` with no context, so a half-open connection blocked `io.Copy` forever and latched update state. The client now uses a transport with dial, TLS-handshake, and response-header timeouts (catching a server that stalls before replying without capping a legitimate large body), and `Download` gets a per-request context with a generous absolute deadline that also covers the body stream. `CheckVersion` keeps its own 30s context. Nothing else uses the client.

- **Unsynchronized reads of `daemonCallbacks.sig` (MEDIUM).** The reconnect path writes `d.sig` under `d.mu`; several reads ran from pion callbacks / FSM goroutines outside the lock (a data race). Added a `currentSig()` accessor that reads under `d.mu` and routed the racy reads (the `OnICECandidate` closures in `webrtc.go` and `voicemail.go`, `OnCallReturn`/`OnCallReturnCancel`, and `publishVoicemailState`/`publishVoicemailStateInitial`) through it. Reads already performed under `d.mu` (conference.go, link-health, the synchronous `InitiateCall`/`AnswerCall`/`HangupCall` sends) were left as-is since they are already synchronized against the same writer.

- **Recovery mount stacking (MEDIUM).** `clearRecoveryFlags` unconditionally mounted `/dev/mmcblk0p4` on `/data` then unmounted once. The try-again exits already run with `/data` mounted, so this stacked a second mount and left the original live. It now detects whether `/data` is already a mount point and operates on the live mount in that case; the cold path still mounts, clears, and unmounts.

- **Mixer ALSA consolidation (LOW, skipped).** `digits-mixer.service` carries a codec-readiness wait loop (`until [ -d /proc/asound/... ]`) that `digitsd.service`'s inline `alsactl restore` ExecStartPre does not, and the unit-enable path is applied outside the working-tree scripts. Folding them risked restoring mixer state before the codec node enumerates, which the finding explicitly warned against, so this was left alone.

- **Unbounded debug PCM capture (LOW).** With `CAPTURE_PCM` set, the render loop wrote every mixed frame including idle silence to disk with no cap (~345 MB/hr) for the daemon's lifetime. Added a size cap (default 512 MB, override with `CAPTURE_PCM_MAX_MB`, 0 = unbounded); once reached the loop stops writing and logs the cutoff once.

- **No-op filter pass (LOW).** `captureLoop` called `p.filters.Process` unconditionally; on the default config `p.filters` is nil and `Process` still allocated and copied a fresh buffer 50x/sec to no-op. Guarded with a nil check. `ExtractChannel` already returns a fresh slice, so the in-place denoiser downstream stays safe.

## Residual risk

The reconnect change shifts the post-reconnect resume/teardown (`tryResumeAfterReconnect`, `TearDownAllMeshPeers`, `HangupCall`, `ctrl.Reset`) from the main goroutine onto the reconnect goroutine. Those touch `ctrl` (mutex-guarded FSM) and `cb` (under `d.mu`), which are already accessed concurrently from pion callbacks, so the locking is correct, but the concurrency window is new and was not exercised against live hardware in this change. The logic and ordering are unchanged from the original inline loop; only the host goroutine differs.

## Test plan

- `make test`: all packages pass.
- `go test -race ./cmd/digitsd/... ./internal/audio/... ./internal/updater/...`: race-clean, including a new concurrent reader/writer test for `currentSig`.
- New tests: updater download success + stalled-body deadline enforcement, mixer capture size cap + unbounded, recovery `/proc/mounts` parsing.
- `golangci-lint run ./...`: 0 issues.
- `make build`: Docker cross-compile to linux/arm64 succeeds.

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -1508,12 +1508,11 @@ func main() {
 			slog.Warn("could not load pairing tones", "error", err)
 		}
 	}
-	mixer.Start()
-	defer mixer.Stop()
-
 	// Debug: capture raw PCM output if CAPTURE_PCM is set. Bounded so a forgotten
 	// capture cannot fill /data (raw 48kHz stereo S16LE is ~345 MB/hr). Default
 	// cap 512 MB; override the megabyte limit with CAPTURE_PCM_MAX_MB (0 = off).
+	// Set up before Start() so the render goroutine observes the capture fields
+	// from its first iteration without any cross-goroutine write to race on.
 	if capPath := os.Getenv("CAPTURE_PCM"); capPath != "" {
 		maxBytes := int64(512) * 1024 * 1024
 		if v := os.Getenv("CAPTURE_PCM_MAX_MB"); v != "" {
@@ -1529,6 +1528,9 @@ func main() {
 			defer mixer.DisableCapture()
 		}
 	}
+
+	mixer.Start()
+	defer mixer.Stop()
 
 	deviceID, err := config.LoadOrCreateDeviceID()
 	if err != nil {
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -121,17 +121,17 @@ type daemonCallbacks struct {
 		candidates []string // local ICE candidates gathered during ring, sent on pickup
 		caller     string   // pendingCaller at time of preparation
 	}
-	iceServers            []owebrtc.ICEServerConfig // cached STUN/TURN servers from signald
-	debugMode             bool                      // read from DIGITS_DEBUG env at startup
-	paired                atomic.Bool
-	pairingCode           string      // current pairing code from server
-	pairingCodeExpiresAt  time.Time   // server-reported expiry of the current pairing code
-	callPeer              string      // number of the remote party during an active call
-	isCaller              bool        // true if we initiated the current call
-	callReturnOrigin      atomic.Bool // true when the current call was initiated via *69
-	isRestartingICE       bool        // true while an ICE restart is in progress
-	restartTimer          *time.Timer // timeout for ICE restart attempt
-	disconnectTimer       *time.Timer // debounce before reacting to pion Disconnected
+	iceServers           []owebrtc.ICEServerConfig // cached STUN/TURN servers from signald
+	debugMode            bool                      // read from DIGITS_DEBUG env at startup
+	paired               atomic.Bool
+	pairingCode          string      // current pairing code from server
+	pairingCodeExpiresAt time.Time   // server-reported expiry of the current pairing code
+	callPeer             string      // number of the remote party during an active call
+	isCaller             bool        // true if we initiated the current call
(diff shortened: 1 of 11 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the review. Addressed in `9a9484c2`.

**#3 (capture-counter race) - fixed.** Real `-race`-detectable race on `captureBytes`/`captureFull`/`captureMax`: `EnableCapture` wrote them under `m.mu` *after* `mixer.Start()` launched the render goroutine, which reads them lock-free. Rather than add locking to the hot render path, I moved the `CAPTURE_PCM` setup ahead of `mixer.Start()`, so the render loop observes the capture fields from its first iteration with a happens-before edge from goroutine creation. As a bonus the defer order now stops the loop fully before the file is closed at shutdown. Updated the `EnableCapture` doc to state the before-Start contract.

**#4 (orphaned `clampAdd` comment) - fixed.** Moved the doc comment back onto `clampAdd`.

**#1 (loop-body sends use the dead `sig` during the reconnect window) - intentional, no change.** During that window there is no live client to reach: the old client is closed and a freshly reconnected one immediately drives call teardown via `tryResumeAfterReconnect`/`TearDownAllMeshPeers`, so a stale-state best-effort signal landing mid-reconnect would be wrong, not useful. The actual win is that events are now *drained* from `sp.Events()` instead of overflowing the 64-deep buffer; whether their best-effort sends land is moot. Routing the seven loop-body sites through `cb.currentSig()` would only close a sub-millisecond window (between the goroutine setting `cb.sig` and the main loop's `<-reconnected` swap) at the cost of churning the reconnect concurrency you just validated, so I'm leaving it as the deliberate callback-vs-loop asymmetry.

**#2 (resume/teardown concurrency) - no change.** This is the hardware-exercise note already called out in the PR body; the lock discipline (`ctrl` mutex + `cb.mu`) is unchanged, only the host goroutine and concurrency window are new.

PR #729, opened 2026-06-09
fix(server): record link health from every signald replica

justinlindh:
## Bug

On `/call/live` screens only one side of a call had telemetry (example: prod call 76297, where the caller logged 7 samples and the callee logged zero despite a 64s call). Prod data shows the pattern exactly: since the week of 2026-05-04 the asymmetric calls are **always caller-only, never callee-only**, and both-sides coverage dropped to ~1/3 of answered calls. That week is when prod moved to k8s with three signald replicas.

## Root cause

`HealthStore` sessions are pod-local memory:

- `Tracker.StartCall` ran `HealthStore.Init(callID)` only on the pod handling the caller's `call` message, which is always the caller's own WebSocket pod.
- `recordSession` silently dropped samples for sessions the local pod never initialized.
- The callee's WebSocket lands on the same pod as the caller's only ~1/3 of the time with three replicas; the other 2/3 of its `link_health` samples were discarded before ever reaching Postgres.
- Same hole for the live view: the SSE stream subscribes to the serving pod's in-memory store, and `Evict` only ran on the pod that handled the hangup, leaking sessions elsewhere.

## Fix

Mirrors the cross-pod pattern the dashboard `events.Broadcaster` already uses:

- **Lazy sessions.** `Record`, `RecordEdge`, and `Subscribe` create the session if the pod hasn't seen it, so each pod ingests and flushes its own phones' samples no matter which replica set the call up. This alone fixes the persisted data.
- **Redis fan-out.** Locally ingested samples and locally triggered evict/disconnect events publish on a `digits:linkhealth` channel; every pod applies the others' events to its rings and live SSE subscribers. Remote-applied samples advance `lastFlushed`, so only the ingesting pod writes to Postgres (the v20 unique indexes remain the backstop). Publishes happen at the public API boundary; session internals stay transport-free.
- **Lifecycle.** Cross-pod evicts close remote subscribers promptly; a 15-minute idle sweep (skips sessions with live subscribers) bounds sessions whose end event predates a pod restart or was recreated by a late sample.

Single-replica deployments are unaffected: with no Redis configured the publishes are no-ops and behavior reduces to lazy creation plus the sweep.

## Tests

- miniredis unit tests: cross-pod sample fan-out (2-party + conference edge), delivery to remote subscribers, evict closing remote subscriptions, disconnect notify, self-event skip.
- Postgres integration test: the remote pod's flusher writes nothing; the ingesting pod writes the row.
- Updated the unit tests that encoded the old drop-when-uninitialized contract, plus new sweep coverage (reaps idle, spares active/subscribed).

## Verification plan

After merge and the ArgoCD rollout, confirm on prod that the next real calls have `call_link_health` rows from **both** endpoints, and spot-check a `/call/live` page.

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -168,6 +168,11 @@ func run(ctx context.Context) error {
 	}
 	healthStore := calls.NewHealthStore(database, healthOpts...)
 	tracker.SetHealthStore(healthStore)
+	if redisBridge != nil {
+		healthStore.SetRedis(redisBridge.Client(), redisBridge.PodID())
+		go healthStore.RunRedis(ctx)
+		slog.Info("redis fan-out enabled for cross-pod link health")
+	}
 
 	healthDone := make(chan struct{})
 	go func() {
--- a/server/internal/calls/linkhealth.go
+++ b/server/internal/calls/linkhealth.go
@@ -84,6 +84,12 @@ type sessionRings struct {
 	mu          sync.Mutex
 	byEndpoint  map[endpointKey]*ring
 	subscribers map[*subscriber]struct{} // nil until first Subscribe
+	// lastActivity is the wall-clock time of the most recent init, sample
+	// append, or subscribe. The idle sweep in Run evicts sessions whose
+	// lastActivity is older than idleSessionTTL and that have no live
+	// subscribers, bounding memory when an end event never reaches this
+	// pod (it may have been handled by a different replica).
+	lastActivity time.Time
 }
 
 type subscriber struct {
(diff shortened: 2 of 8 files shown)


justinlindh:
@claude review

justinlindh:
Addressed in e642937f:

**Finding 1 (subscribe-after-end phantom stream):** fixed both ways. The 2-party and conference SSE handlers now re-check liveness from the DB immediately after subscribing and flip straight to the terminal `ended` state when the call ended in the race window. As a self-heal for any other path to a phantom session, the sweep now reaps sessions with live subscribers after `subscribedSessionTTL` (1 hour of total sample silence), closing their streams with `EndedKind`. Added `TestHealthStoreSweepReapsStuckSubscribedSession` covering exactly the scenario you described (subscribe to a call whose evict already passed this pod), plus the assertion that the normal idle TTL still spares subscribed sessions.

**Finding 2 (failover dual-ingest):** documented on the invariant comment in `recordSession`: a WebSocket failover can briefly produce two publishers for one endpoint; ON CONFLICT swallows the duplicate direction and a skipped write in that window is one sample, recovered on the next flush tick.

**Finding 4 (disconnect asymmetry):** added the one-liner on `notifyDisconnectedSession` explaining why it stays drop-if-absent while Record/Subscribe create-if-absent.

Finding 3 (memory scales with replicas x active calls) acknowledged as intended: bounded by RingCapacity and the sweep, fine at this fleet's size. The `context.Background()` publish nit matches the broadcaster's convention, leaving as is.

Full integration suite and lint green.

PR #735, opened 2026-06-10
refactor(firmware): remove deprecated no-op tone module

justinlindh:
## What

Deletes `tone.c` / `tone.h` and every call into them. The module has been a set of empty stubs since tone generation moved to the Pi side; its own header says the stubs exist only "so the FSM compiles without changes." This pays that debt instead of carrying it:

- `phone_fsm.c`: drop the `#include`, the 14 `tone_play` / `tone_stop` / `tone_play_dtmf` / `tone_update` calls scattered through `set_state`, `process_key`, and the update loop, and the `TONE:DIAL` / `TONE:RINGBACK` / `TONE:STOP` UART command branches that only invoked no-ops.
- `main.c`: drop `tone_init()`.
- `CMakeLists.txt`: remove the source file.
- `docs/architecture/uart-protocol.md`: remove the three tone commands from the Pi-to-Pico table and state that tones are Pi-side only.

## Why the wire contract is unchanged

The removed UART branches sent no response and had no effect, and the firmware's command dispatcher ignores unrecognized commands silently. digitsd still emits one `TONE:DIAL` (post-pairing restart path in `cmd/digitsd/main.go`); the Pico ignored it before as a no-op and ignores it now as an unknown command, byte-for-byte the same on the wire. That sender is left alone here because deciding whether the post-pairing dial tone should be re-routed to the Pi mixer is a digitsd behavior question, not firmware cleanup.

## Why this scope

Smaller (stubs stay, call sites stay) keeps dead indirection in the hottest file of the firmware and keeps the protocol doc lying about supported commands. Bigger (also touching digitsd) would mix a behavior decision into a behavior-neutral deletion.

## Validation

Full firmware build against Pico SDK 2.2.0 (the version pinned in `Dockerfile.builder`) compiles and links clean, UF2 produced.


justinlindh:
Resolved merge conflicts from main (#742 landed there).

**How it resolved:**
- `tone.c` / `tone.h`: delete/modify conflicts resolved in favor of deletion. Deleting the files subsumes #742's em-dash comment fixes in them, so nothing of value was lost.
- `phone_fsm.c`, `main.c`, `firmware/CMakeLists.txt`: auto-merged cleanly. The PR's tone removals (include, 14 `tone_*` calls, the `TONE:DIAL`/`TONE:RINGBACK`/`TONE:STOP` UART branches) applied on top of main's current versions (hoisted includes, removed dead exports, `-Wall -Wextra`). The result keeps main's `-Wall -Wextra` and drops `tone.c` from sources.
- `firmware/test/`: delinked `tone.c` from the host-test `CMakeLists.txt` link list and updated `README.md` ("real `ringer.c` links directly"). No test asserted behavior of the removed `TONE:` branches, so no test logic changed: unknown commands are already ignored silently by `process_pi_command`.
- `docs/architecture/uart-protocol.md`: the PR's doc change stands (main did not touch this file).
- digitsd's `TONE:DIAL` sender stays as-is per the wire-contract argument; the firmware now ignores those commands harmlessly.

**Verification:**
- Host tests: `make test` in `firmware/` passes, 35 tests / 132 checks / 0 failures (unchanged count).
- Full firmware build: clean rebuild via `scripts/build.sh` (SDK at `/usr/share/pico-sdk`), zero warnings under `-Wall -Wextra`.
- grep: zero references to `tone_play`/`tone_stop`/`tone_play_dtmf`/`tone_update`/`tone_init`/`tone.h` anywhere in `firmware/`, `pi/`, or `server/`.

PR #738, opened 2026-06-10
refactor(server): assert paired state through a test-local helper

justinlindh:
## What

The dead `pairing.Store.IsPaired` removal already landed on main via #739, so there is nothing left for this PR to drop from production code. What #739 did not preserve were the stronger paired-state assertions and the generate-does-not-pair property; its simpler deletion removed them along with the method's own tests. This PR restores those, test-only, on top of the current main.

## How the tests were untangled

The paired/unpaired assertions are still worth making, they just are not assertions about a production API anymore:

- A small `isPaired` helper (a `paired_at` readback) lives in `pairing_test.go`, shared by the unit and e2e integration tests in the package. Its comment records why production has no such reader: the device learns it is paired from the claim response pushed over its WebSocket, so the server never answers "is this hardware ID paired?" at runtime.
- `TestClaimDevice_Success` and the e2e flow assert paired state through the helper.
- `TestGenerateCode_DoesNotPair` guards the production property that generating a code must not set `paired_at`.
- `TestIsPaired_FalseForUnknown` stays deleted: "unknown hardware ID returns false" was a property of the removed method, and keeping it would only test the helper.

## Validation

`go build ./...`, `go vet ./...`, and the unit test tier (`make test`) pass. The pairing integration tests (`go test -tags=integration ./internal/pairing/...`) were run against a disposable Postgres and pass, including the restored assertions. `golangci-lint run ./...` reports zero issues, and the pairing package is clean under integration tags as well.


justinlindh:
Resolved the merge conflict with main. The core change this PR originally proposed (dropping `pairing.Store.IsPaired`) already landed via #739, so I took main's `pairing.go` verbatim and reduced this PR to the surviving test improvements that #739's simpler deletion left behind:

- the test-local `isPaired` helper (the `paired_at` readback, with the comment recording why production has no such reader),
- the paired-state assertion in `TestClaimDevice_Success` and the e2e flow through that helper,
- `TestGenerateCode_DoesNotPair`, guarding that generating a code must not set `paired_at`.

`TestIsPaired_FalseForUnknown` stays deleted (it would only test the helper). e2e step numbering was reconciled to stay coherent (1 through 10) after re-inserting the paired-state checks.

Retitled to match the actual content and updated the body. Verified: `go build`, `go vet`, `make test`, the pairing integration suite against a disposable Postgres (all pass), and `golangci-lint run ./...` clean.

PR #739, opened 2026-06-10
refactor(server): remove dead pairing accessor and add migration idempotency test

justinlindh:
Three small server-side quality fixes.

## 1. Remove dead `Store.IsPaired`

`pairing.Store.IsPaired` had no production caller. A repo-wide grep (`grep -rn "IsPaired" --include="*.go" server/`) found it referenced only in test files and its own definition: nothing in the runtime pairing flow (`GenerateCode`, `ClaimDevice`, `ClaimDeviceToLine`, `CleanupExpired`) or any handler calls it. Removing it is safe.

The IsPaired-specific test assertions are removed while the surrounding tests stay intact:
- `pairing_test.go`: `TestClaimDevice_Success` keeps its token-length assertion; the secondary IsPaired check is dropped. The two IsPaired-only tests (`TestIsPaired_FalseForUnknown`, `TestIsPaired_FalseBeforeClaim`) are deleted.
- `e2e_pairing_test.go`: the IsPaired before/after/both verification steps are dropped; the claim is still proven by the token and hardware-id assertions, and the code-reuse, duplicate-number, second-claim, and household-membership steps are unchanged. Steps renumbered to stay sequential.

## 2. Migration idempotency test

`internal/db` had no test file. Added `db_integration_test.go` (integration build tag, `TEST_DATABASE_URL` skip convention matching the other integration tests). It opens a database (initial migrate), then opens it again on the same DB and asserts no error. The second `Open` re-runs every migration statement over an already-migrated schema, which is the exact production boot path on restart.

This locks the contract the whole migrate mechanism rests on: `migrate()` runs every statement unconditionally on each boot, relying on `IF NOT EXISTS` guards and `schema_version` gates to be a no-op on an already-migrated database. A statement that silently lost its guard would now fail this test instead of crashing a production process on restart.

## 3. CLAUDE.md correction

The repo-root `CLAUDE.md` described the server web UI as "htmx + Tailwind". Tailwind appears nowhere in `server/`: no config, no CDN link in any template, no utility-class usage (`grep -rni "tailwind" server/internal/web/` is empty). The UI ships custom CSS: `digits.css` plus per-theme `dialup.css` and `answering-machine.css` in `internal/web/static/`, linked directly from the `layout-*.html` templates. Both CLAUDE.md references corrected to say custom CSS.

## Verification

- `go build ./...`: clean.
- `go vet ./...` and `go vet -tags integration ./internal/db/ ./internal/pairing/`: clean.
- `make test` (unit tier): all packages pass.
- New test against disposable Postgres: `TestMigrateIsIdempotent` PASS. Full pairing integration suite PASS with the IsPaired tests gone.
- `golangci-lint run ./...` and `golangci-lint run --build-tags=integration ./internal/db/ ./internal/pairing/`: 0 issues.


justinlindh:
@claude

justinlindh:
Thanks for the review. Both nits addressed:

**Em-dash nit (fixed in 482be2c9).** You're right, the renumber pulled three pre-existing em-dash step comments into the diff (`e2e_pairing_test.go` Steps 3, 5, 6). My earlier local sweep used a PCRE byte pattern that silently failed to match U+2014, so I'd misread them as clean; a `LC_ALL=C` byte grep confirmed your finding. Replaced each em dash with a natural-reading clause (`Reuse same code, which should fail`, etc.) so the file is now em-dash-free.

**Stale PR description (fixed).** The description still claimed a third explicit `migrate()` call. A simplify pass on this branch had already dropped that redundant pass (the second `Open` is the full production boot path and the only assertion the contract needs), so the code was correct and only the prose lagged. Updated the description to match: two `Open` calls, no third migrate.

No findings skipped. Re-verified after the change: `go build ./...`, `go vet -tags integration ./internal/pairing/`, `go test ./internal/pairing/`, and `golangci-lint run --build-tags=integration ./internal/pairing/` all clean.

PR #740, opened 2026-06-10
refactor(image): shell quality pass and shellcheck CI across image, tools, and scripts

justinlindh:
A quality pass over the shell/ops scripts in `pi/image/`, `tools/`, and `scripts/`: dead-code deletions, stale metadata fixes, DRY consolidation scoped by runtime context, and a new shellcheck CI gate. No behavior change for any successful build path; every change is internal restructuring or correctness of comments/metadata, hence `refactor`.

## Deleted (each verified unreferenced via repo-wide grep)

- **`digits-mixer.service`**: never enabled (no `WantedBy` symlink shipped, no `hostside_enable_service` call) and never started. Its only job, `alsactl restore ... /data/digits_mixer.state`, is already done by `digitsd.service` `ExecStartPre`. The assets test comment that cited it as a stale-file example is reworded to a generic case so the test stays meaningful.
- **`wpa_supplicant@wlan0.service.d/digits.conf`**: a pre-NetworkManager provisioning drop-in pointing `wpa_supplicant` at `/data/wifi/wpa_supplicant.conf`, a file nothing writes anymore. The remaining `wpa_supplicant` references (`verify.go`, `digits-ap-check`) only `systemctl stop` the unit, which does not need the drop-in. Removing the file leaves the drop-in directory empty and git drops it.
- **`init-data.sh` `--pcb` no-op**: `PCB_MODE` was parsed then never read (shellcheck SC2034), so the flag did nothing. Removed the parsing block, fixed the usage strings, and dropped the single caller that passed it (`build-image.sh`'s `init-data` invocation). `build-image.sh`'s own `--pcb` flag is real (V1/V2 image naming, codec overlay, mixer-state selection, SWD config) and is untouched.

## Stale metadata and comments

- `Documentation=` in five units (`digits-ap-check`, `digits-ap`, `digits-dnsmasq-ap`, `digits-first-boot`, `digits-setup`) pointed at `github.com/digits/digits`, a nonexistent repo. Repointed to `github.com/justinlindh/digits` and added the same line to `digitsd.service`.
- `cmdline-ro.txt.fragment` said `/data` is `p3`; it is `p4` (`p3` is recovery).
- `fstab.fragment` cited `etc/systemd/system/data-*.mount` unit files that do not exist; the bind mounts are declared inline via `x-systemd.requires=/data`. Removed the dead reference.
- `init-data.sh` described `wifi/` as holding a setup-portal-written `wpa_supplicant.conf`; provisioning moved to NetworkManager and the dir now holds `.nmconnection` backups managed by digitsd (`configure.go` `backupDir = /data/wifi`). Reworded both mentions.

## Consolidation map (scoped by runtime filesystem context)

The build scripts run in two different places, so consolidation is deliberately partial.

**Shared** via new `pi/image/lib/log.sh` (die/info/warn), sourced by the four scripts that always run with the repo tree present:

| Script | Context |
|---|---|
| `tools/build-image.sh` | host and inside the Docker image-builder (repo bind-mounted at `/digits`) |
| `pi/image/init-data.sh` | same; invoked by build-image.sh via absolute path |
| `pi/image/partition-setup.sh` | same |
| `pi/image/build-docker.sh` | host only |

Each resolves the lib via its own `SCRIPT_DIR`/`REPO_DIR`, valid in both host and container roots. Sourcing it into `build-docker.sh` also fixes a latent bug: it called `warn` on the missing-`GITHUB_TOKEN` path but never defined it.

**Left duplicated on purpose:** `pi/image/entrypoint.sh` keeps its local `die`/`info`. It is `COPY`ed into the Docker image as `/entrypoint.sh` and runs as the build bootstrap, so it cannot assume the shared lib is reachable at parse time. A comment in both `entrypoint.sh` and `lib/log.sh` records the intent.

**Other unifications:**
- `tools/build-firmware.sh` used an inline `sha256sum | awk`; it now uses the same `sha_file()` helper shape as `tools/build-pi.sh`. Output (`Built:` / `SHA256:` lines) is unchanged.
- `pi/image/build-docker.sh` arg parsing converted from `for`+`case` to the `while`+`case`+`shift` pattern used by `build-image.sh`, `entrypoint.sh`, and `init-data.sh`. Behavior verified identical for flags-then-image, image-only, and no-args.
- `require_cmd` was defined byte-identically in `tools/build-image.sh` and `pi/image/partition-setup.sh` and depends on `die`, which already lives in the shared lib. Moved it into `pi/image/lib/log.sh` and deleted both local copies; call sites unchanged. The lib filename stays `log.sh` to avoid churning the `source=` directives.

## CI gate

New `.github/workflows/shellcheck-ci.yml` lints every shell script under `tools/`, `scripts/`, and `pi/image/`, including the no-extension executables in `rootfs-overlay` (selected by shebang; the lone Python script `digits-pico-monitor` is excluded). It runs at **`-S warning`** (warnings and errors fail; pre-existing info-level findings in untouched scripts are allowed) and **`-x`** so the shared `lib/log.sh` source is followed instead of tripping SC1091. The workflow is path-filtered to those three trees plus its own file, so it only runs when relevant scripts change.

Findings fixed so the gate passes clean on this branch:
- `scripts/flash.sh`: `2>/dev/null` sat between `find`'s path list and its predicates (SC2227); moved the redirect to the end.
- `digits-ap-check`: the `nullglob` array build is an intentional glob (SC2206); narrow disable with a reason.
- `initramfs/recovery-root.sh`: `rootmnt`/`ROOT` are injected by initramfs-tools, not assigned in the script (SC2154); documented disable.
- `build-image.sh` kernel-module copy: two pre-existing warnings (SC2043 intentional single-entry loop, SC2010 well-known kernel dir names) get narrow reasoned `disable` directives rather than a risky rewrite of working image logic.

## Local pre-commit gate

The repo's pre-commit hook (`core.hooksPath .githooks`) already linted staged Go files. It now also shellchecks the shell files a commit touches, mirroring the CI gate: same `-S warning -x` severity and the same `tools/`, `scripts/`, `pi/image/` scope (rootfs-overlay executables selected by shebang, so the Python script is skipped). It checks only staged files, runs in well under a second, and skips with a notice rather than blocking when `shellcheck` is not installed, matching how the existing `golangci-lint` step handles a missing tool. This catches a shellcheck regression locally instead of waiting for CI.

## Verification evidence

- shellcheck `-S warning -x` over the full 23-file workflow set (now including `pi/image/lib/log.sh` directly): exit 0.
- `bash -n` / `sh -n` on every modified script, including the pre-commit hook: pass.
- `cd pi/digitsd && make embed && go build ./... && go test ./internal/assets/...`: build OK, assets tests `ok` (the overlay embed test passes after the unit deletion).
- Pre-commit hook exercised against fabricated staged sets: clean shell files pass, a warning-level finding blocks (exit 1) at `-S warning`, the Python monitor and non-shell files are excluded, and a missing `shellcheck` skips with a notice instead of blocking.
- `require_cmd` now resolves only from the shared lib; both call sites (`build-image.sh`, `partition-setup.sh`) source it first and still pass `bash -n` and shellcheck.
- Traced `build-image.sh` end-to-end in the Docker context: `SCRIPT_DIR=/digits/tools`, `REPO_DIR=/digits`, sources `/digits/pi/image/lib/log.sh`; the `partition-setup.sh` and `init-data.sh` sub-invocations each resolve the same shared lib via `SCRIPT_DIR=/digits/pi/image`. No sourcing path breaks. No full image build was run.

## Deliberate skip

`docs/build/wiring.md` still lists `digits-mixer.service` as Active. It is outside this pass's scope (the three named script directories, not `docs/`) and is broadly older hardware-wiring notes (it also references `/home/digits/restore_mixer.sh` and a `/home/digits/...` mixer-state path from the V1-HAT era). Fixing one row there would be scope creep into a doc that needs its own freshness pass; left as-is to keep this PR focused.


justinlindh:
@claude

justinlindh:
Thanks for the review. Both findings verified against the code:

**1. `lib/log.sh` not linted directly.** Correct, and worth fixing. The `mapfile` glob had `pi/image/*.sh` but not `pi/image/lib/*.sh`, so `log.sh` was only checked transitively via `-x`. Fixed in bc8c8906: added `pi/image/lib/*.sh` to the explicit list. The gate now lints 23 files including `log.sh` as a standalone target (it already carried `# shellcheck shell=bash` and passes), so the "lints every shell script under pi/image/" claim is now literally true.

**2. `build-docker.sh` flags-before-positional.** Correct call, no code change. The `for arg in "$@"` to `while [[ "${1:-}" == --* ]]` switch was the intended consistency change (task asked to match `build-image.sh`, `entrypoint.sh`, `init-data.sh`, all of which stop at the first non-flag). As you confirmed, every caller (`Makefile`, `README-image-builder.md`, the documented `[--dev] [--pcb] [image]` synopsis) uses flags-first, so the new behavior is identical for all real orderings. Restoring order-independence no caller exercises would just re-diverge from the siblings, so I'm leaving it. The "behavior verified identical" claim is scoped to caller orderings, as you noted.

The `docs/build/wiring.md` skip stands for the reason in the PR body.

justinlindh:
Added two low-effort items from the tools/scripts re-assessment:

- **28a2aef1**: extracted the byte-identical `require_cmd` from `build-image.sh` and `partition-setup.sh` into the shared `pi/image/lib/log.sh` (it depends on `die`, already there). Call sites unchanged; both still pass `bash -n` and shellcheck. Kept the filename as `log.sh` to avoid churning the `source=` directives.
- **2cbfd466**: the pre-commit hook (`core.hooksPath .githooks`) now shellchecks the shell files a commit touches, mirroring the CI gate (`-S warning -x`, same tools/scripts/pi/image scope, rootfs-overlay by shebang). Staged-only, sub-second, and skips with a notice if `shellcheck` is absent so it never blocks on a missing tool.

Left untouched per the assessment: SCRIPT_DIR boilerplate, entrypoint.sh local helper copies, and pi/swd/flash-pico.sh.

PR #741, opened 2026-06-10
refactor(digitsd): extract signaling dispatch from main and prune stale debug tools

justinlindh:
## Part 1: dispatch extraction

main()'s event loop carried a ~500-line inline `switch msg.Type` that routed every inbound signaling message. That switch is now `func (d *daemonCallbacks) handleSignal(msg *signal.Message)` in `pi/digitsd/cmd/digitsd/dispatch.go`. The loop case shrank to:

```
case msg := <-sig.Inbox():
	cb.handleSignal(msg)
```

This is a mechanical move, not a redesign. All 27 case labels were verified identical in order and set against the pre-change `main.go` (diff of the case-label list is empty), and the single outer `default` is preserved. Same locking (`d.mu`), same goroutine spawns, same log lines. `main.go` is 474 lines shorter net.

### Loop-local dependencies

Handlers reference state the run loop owns. Those that are set-once and stable for the daemon's life are now fields on `daemonCallbacks`, wired in one block right before the loop starts: `deviceID`, `serverURL`, `flashCapable`, `requeryFirmware`, `contactsCache`, `pairingRefresh`. `flashCapable` (`*atomic.Bool`) and `pairingRefresh` (`*time.Timer`) are shared by identity, so a `Store`/`Reset` in either the loop or a handler is seen by the other. Firmware version is read through the existing synchronized `getFirmwareVersion()` getter rather than a captured local.

### Testability

The controller is reached through a new `signalController` interface (the 11 `*phone.Controller` methods the dispatch calls). The running daemon sets `cb.ctrlSignal = ctrl`; tests inject a recording fake. `main.go`'s own `cb.ctrl` usage is unchanged.

`dispatch_test.go` adds table-driven coverage:

- Routing for every FSM-delegating type (ring, answer, hangup, busy, the five conference types, both call-return ring/result, and error-in-ADD_CALLING), asserting the right controller method fires with the right args.
- Cache/config-mutating cases through their observable effects: contacts (populated installs a checker, empty clears it, both message types), ice-servers (list cached with creds), line_settings (voice style applied and persisted to disk; nil payload is a no-op).
- The unknown-type default (no controller method fires) and the DTMF state guard (digit while not connected is dropped).

## Part 2: stale debug tools

Grepped the whole repo (docs, Makefiles, scripts, CI, Go) for each of the seven `cmd/` debug tools.

| Tool | Decision | Rationale |
|---|---|---|
| `memprofile` | keep | Per-call memory calibration harness for pion peer connections on the Pi Zero 2 W. Named ongoing value. |
| `alsatest` | keep | First-line "is the codec making sound at all" check during V1/V2 codec bring-up. 54 lines, near-zero maintenance. |
| `dmixlat` | keep | Measures actual on-device ALSA output latency via `snd_pcm_delay()` (cgo). Hardware-level buffer-depth check not duplicated elsewhere. |
| `latclient` | keep | Structurally coupled to the daemon's Unix socket server, which auto-answers calls from this client (referenced by name in `main.go` and `socketserver.go`). |
| `latbench` | delete | Synthetic-sine Opus encode/decode/playback microbench. Coverage now lives in `internal/codec` unit tests and `latclient`'s real-call timing. |
| `pipetest` | delete | 47-line synthetic-sine Opus throughput loop, no hardware. Same coverage as above. |
| `clocksync` | delete | Two-host UDP clock-offset one-off needing a partner host. No current workflow. |

The only repo references to all seven were rows in `pi/digitsd/README.md`'s table; that table now lists the four keepers, and `cmd/README.md` documents what each keeper is for. Deleted code lives in git history.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` from `pi/digitsd`: all pass. `go test -race ./cmd/digitsd/` clean.
- `golangci-lint run ./...` (v2.11.4, repo `.golangci.yml`): `0 issues`.
- Cross-compile: `make build` (Docker, linux/arm64, cgo with libopus + libasound) exits 0.
- Behavior parity: case-label list diff between pre- and post-change is empty (27 labels, same order, one default).
- No build tags or cgo touched; the kept cgo tool (`dmixlat`) is unmodified.


justinlindh:
@claude

justinlindh:
Thanks for the review. Addressed below; pushed b1e8f41f.

**1. `TypeUpdateTrigger` firmware version: this one is actually a pure move, not a behavior change.** I re-diffed the pre-refactor case body. The original did not read a startup-captured value: it read the loop-local `fwVersion`, which the `<-fwVersionCh` case reassigned on every firmware-version event (old `main.go:2229`), and that same case also called `cb.setFirmwareVersion(...)` in lockstep. So the loop-local and the struct field are written together at a single site and cannot diverge; both are seeded to the same startup value at construction. `d.getFirmwareVersion()` returns exactly what the old loop-local held at the call site. No fresher-vs-staler difference is introduced, so the `refactor` typing is correct and nothing here warrants a `fix`. Good catch to check it though, the framing in my PR body ("read through the synchronized getter") undersold that the original was already live, not captured.

**2. Inconsistent mixer access: fixed in b1e8f41f.** Routed every arm through the `mixer := d.mixer` alias declared at method entry. Same object, no functional change, reads uniformly now.

**3. `newDispatchDaemon` nil fields: skipping, by design (YAGNI).** Pre-populating `sig`/`pairingRefresh`/`flashCapable` with inert stubs would be defensive scaffolding for tests that don't exist. The currently-tested message types avoid those paths by construction (the busy-retry send sits behind a `State()` guard the fake leaves non-CALLING; no tested type touches `pairingRefresh.Reset` or `flashCapable.Load`). Whoever adds a case that needs them wires the one field that case needs, which is clearer at the call site than a fully-populated shared fixture. Noting it in the helper would just restate what the per-test construction already shows.

PR #742, opened 2026-06-10
refactor(firmware): add host test harness and remove dead code

justinlindh:
## What and why

Two-part firmware quality pass.

**Host test harness.** The pure-logic firmware modules (hook, keypad, uart_proto, phone_fsm) had no automated coverage: every behavioral edge case lived only in comments and on-device testing. This adds a native unit-test setup so those modules run and get asserted off the device, with a deterministic virtual clock and GPIO so timing windows are reproducible.

Production sources are compiled unmodified against fake Pico SDK headers under `firmware/test/fakes/` (`pico/...`, `hardware/...`). Those shims forward to a controllable monotonic clock and a GPIO array, plus a fake UART with a captured RX queue and TX buffer. The hardware-bound modules the FSM depends on (board profile, LED, phase) have in-memory fakes that expose state for assertions; the real `ringer.c` and `tone.c` link directly. `test_fsm.c` includes `phone_fsm.c` directly rather than linking it separately so the file-static `buf_appendf` accumulator and `s_state` are reachable.

The harness is dependency-free: a small header with `CHECK` / `CHECK_EQ` / `CHECK_STREQ` and a tiny runner. No Pico SDK and no ARM toolchain are needed, so it runs fast locally and stays cheap in CI. Entry point is `make test` from `firmware/`.

**Cleanup.** Dead-code removal and build hygiene on the production sources, with the firmware target now compiled under `-Wall -Wextra` (clean, no source changes needed to get there).

## Test coverage

35 tests, 132 checks, all passing.

- `hook.c`: 50ms debounce stability window; the flash-vs-hangup classifier over the `[100, 600]`ms window; sub-100ms bounce suppression; past-600ms timeout committing to a real hangup; flash-disable mid-window committing to on-hook; polarity inversion re-sync; software force/release re-sync.
- `keypad.c`: distinct-key acceptance fires exactly once per press; same-key repeat suppressed while held; a same-key mid-press bounce (sub-80ms open that re-closes) produces no second accept; a re-press is accepted only after a confirmed release; a distinct key arriving inside the 80ms window is rejected until it elapses.
- `uart_proto.c`: RX line framing across fragments; CR/LF and blank-line handling; the 127-byte max line; oversized-line overflow discarded then resynced on the next terminator so it cannot corrupt the following line; inject-takes-priority.
- `phone_fsm.c`: the state transition table (idle, dial-tone, dialing, ringing, connected, busy) driven by hook events and Pi commands; the DIALING off-hook timeout firing `TIMEOUT:DIAL_TONE` into BUSY; a full 7-digit dial emitting exactly one `DIAL:<digits>` line with the `s_dial_sent` latch suppressing duplicates; the `buf_appendf` saturating accumulator (truncation boundary, no `size_t` underflow when full, zero-size no-op).

A path-filtered `fw-ci.yml` runs the suite on `firmware/**` changes only, so nothing runs unless the firmware moved.

## Cleanup details

- Removed exported functions with zero callers repo-wide: `phone_fsm_reset`, `phone_fsm_set_keytest`, `phone_fsm_is_keytest`, `ringer_is_active`, `hook_is_flash_enabled` (plus header declarations). `phone_fsm_get_state` and `hook_is_inverted` are kept on purpose: they are the natural state-inspection seams the host tests assert through.
- Removed the write-only `s_start_ms` in `ringer.c`; `s_seg_start_ms` is now seeded from `now_ms()` directly.
- Unified the `led` / `ringer` / `phase` / `tone` header guards to the `DIGITS_*_H` style used by the rest of the headers.
- Moved the stranded SDK `#include` directives in `phone_fsm.c` into the top include block.
- Reworded comment em dashes to colons or periods.
- Added `-Wall -Wextra` to the firmware target.

## Build verification

- Host tests: `make test` in `firmware/` builds clean (`-Wall -Wextra -Wno-format`) and reports `35 test(s), 132 check(s), 0 failure(s)`.
- Full firmware: `./scripts/build.sh` against the Pico SDK builds to `digits.uf2` with zero warnings from `firmware/src/` under the new `-Wall -Wextra`.
- `grep` for em dashes across `firmware/` returns nothing.


justinlindh:
@claude

justinlindh:
Thanks for the review. Both notes addressed:

1. **Stale check count.** Updated the PR description: 113 to 114 checks in both spots, to match the keypad-alignment commit that added a scan assertion.

2. **Doc drift in `fake_env.h`** (fixed in 9266a9e1). Dropped the "every GPIO pull cleared" claim from the `fake_env_reset` doc; the fake tracks only a single level array, no pull state. Reworded the `fake_gpio_get_level` comment to describe the shared level-array readback rather than the unused ringer-H-bridge-inspection use case.

Verified: `make test` is green at `32 test(s), 114 check(s), 0 failure(s)` and the on-device `./scripts/build.sh` builds clean under `-Wall -Wextra` with zero warnings from `firmware/src/`.

justinlindh:
Added the three flagged coverage gaps in c186f08b:

- **DIALING off-hook timeout** (`test_fsm_dialing_timeout_to_busy`): enter DIALING with a partial number, advance 15s, assert `TIMEOUT:DIAL_TONE` fires once and the FSM lands in BUSY. This was the newest merged-in branch with zero coverage.
- **Full 7-digit dial** (`test_fsm_full_dial_emits_single_dial_line`): drive seven digits through `process_key` and assert exactly one `DIAL:<digits>` line is emitted, with the `s_dial_sent` latch suppressing a duplicate on a further keypress.
- **Keypad same-key mid-press bounce** (`test_keypad_same_key_midpress_bounce_rejected`): a brief sub-80ms open that re-closes the same key does not confirm the release, so the re-close is a held no-op with no second accept.

The fake UART now captures TX so emitted lines can be asserted; previously TX was a sink. Counts: `35 test(s), 132 check(s), 0 failure(s)`. `./scripts/build.sh` builds clean under `-Wall -Wextra` with zero `firmware/src/` warnings. ringer.c left uncovered per the YAGNI assessment.

PR #745, opened 2026-06-11
fix(digitsd): resync Pico state after runtime firmware flash

justinlindh:
## The bug

When digitsd's runtime auto-updater applies a firmware-only update (`release_available` -> updater flashes the Pico via `flash-pico.sh`), the Pico reboots. Its firmware boots into `PHASE_PAIRED`, which `firmware/src/main.c` maps to `LED_MODE_BREATHING` at boot by design, expecting the Pi to clear it.

The STARTUP path clears it (`cmd/digitsd/main.go`): after POST it runs `resetPicoHardware(sp)` then `StateSet("PAIRED")`/`("UNPAIRED")`. But the RUNTIME updater path never re-synced Pico state after a flash. The post-flash VERSION re-query (the `requeryFirmware` closure feeding `fwVersionCh`) updated the firmware version and sent `device_info`, then stopped. No `STATE:SET`, no LED command.

## Production evidence

Last night the whole fleet took a firmware-only OTA (`fw/v1.14.1`, with pi already current at `1.42.11`). The journal on every phone showed, after `Flash complete`: a successful `VERSION` re-query, a `device_info` send to signald, and then nothing serial-bound. No `STATE:SET` and no LED command were ever emitted, so every phone stayed on the breathing idle LED. Previous firmware OTAs always rode along with a pi update whose daemon restart cleared the LED via the startup path, which is why this never surfaced before.

## The fix

Re-sync Pico state in the `fwVersionCh` consumer, the existing post-flash VERSION re-query handler. A successful re-query already proves the Pico finished rebooting and is answering UART again, which is exactly the liveness signal the resync needs. It now runs `resetPicoHardware` + `StateSet`, deriving the phase the same way startup does (`DeviceToken == "" ? "UNPAIRED" : "PAIRED"`) through a shared `picoStateForToken` helper so the two paths stay in lockstep.

Seam choice: the resync lives in `cmd/digitsd` where the serial port and live config are in scope, not in `internal/updater`. The updater already signals post-flash liveness through its existing `afterFirmwareUpdated` -> `requeryFirmware` -> `fwVersionCh` chain, so no new wiring was needed and the updater package stays free of serial dependencies.

Scope and safety:

- The resync runs on every Pico reboot re-query, not only firmware OTAs. That is intentional and matches startup, which resyncs unconditionally after POST: any reboot (runtime flash, external SWD, power cycle) leaves the chip breathing and needs the same clear. It is harmless on the pi-binary-only update path because that path restarts the daemon and the startup resync covers it.
- It skips while a call is active (`callPeer != ""`) so a mid-call Pico power cycle cannot stomp the call LED. The OTA flash path is already idle-gated by `runAutoUpdate` (which defers when in a call), so this guard is non-redundant only for the external-reboot case and does not duplicate the OTA gate.
- No recovery/setup regression: the resync only runs in normal mode (the other modes do not consume `fwVersionCh`), and it derives the phase from the device token exactly as normal-mode startup does.

## Test coverage

`cmd/digitsd/postflash_resync_test.go` fakes the serial boundary (the same recording-fake pattern as `linkhealth_test.go`) and asserts:

- `picoStateForToken` maps empty token to `UNPAIRED` and a non-empty token to `PAIRED`.
- `resyncPicoState` emits exactly `STOP_RING`, `LED:UNLOCK`, `STATE:SET:<state>` for both pairing states.
- `resetPicoHardware` clears ring and LED and never emits `STATE:SET`.

## Why fix-typed

The device behavior changes: phones that were stuck breathing now return to the correct idle LED after a firmware OTA. Fix-typed also intentionally triggers a pi release; that release rolling out to the fleet is itself the remediation that clears the currently-stuck LEDs on remote devices.

## Verification

From `pi/digitsd`, all green:

- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `golangci-lint run ./...` (0 issues)
- `go test -race ./cmd/digitsd/`

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -854,8 +854,8 @@ type picoStateResyncer interface {
 }
 
 // resetPicoHardware clears any residual ring or LED state on the Pico in case
-// the Pi rebooted mid-call. Safe no-op on clean boots where none of these
-// hardware states were active.
+// the Pi or Pico rebooted mid-call. Safe no-op on clean boots where none of
+// these hardware states were active.
 func resetPicoHardware(sp picoStateResyncer) {
 	slog.Info("pico: clearing residual hardware state")
 	sp.StopRing()
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -843,13 +843,46 @@ func runTargetedUpdate(serverURL, piVersion, fwVersion, targetPi, targetFW strin
 	}
 }
 
+// picoStateResyncer is the subset of *phone.SerialPort that the Pico
+// hardware-reset and state-resync helpers need. Declaring it as an interface
+// lets tests fake the serial boundary and record the commands emitted, the same
+// way linkhealth fakes its sigSender.
+type picoStateResyncer interface {
+	StopRing()
+	LED(mode string)
+	StateSet(state string)
+}
+
 // resetPicoHardware clears any residual ring or LED state on the Pico in case
 // the Pi rebooted mid-call. Safe no-op on clean boots where none of these
 // hardware states were active.
-func resetPicoHardware(sp *phone.SerialPort) {
-	slog.Info("pico: clearing residual hardware state on startup")
+func resetPicoHardware(sp picoStateResyncer) {
+	slog.Info("pico: clearing residual hardware state")
 	sp.StopRing()
 	sp.LED("UNLOCK")
+}
+
+// picoStateForToken derives the persisted Pico phase from the pairing state the
+// same way the startup path does: an empty device token means the phone is not
+// yet paired. Kept as a pure function so the mapping has a single source of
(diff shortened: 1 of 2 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the review. Disposition of the three observations:

1. **In-call skip guard untested.** Skipping per your own YAGNI call. The guard is a three-line read of `cb.callPeer` under `cb.mu` inline in `main()`; extracting it solely to cover it would add a seam that earns nothing. The logic reads correctly and the surrounding mutex discipline matches every other `callPeer` access.

2. **Reboot during active pairing.** Skipping: benign and pre-existing. A reboot loses the pairing-code display regardless of this resync, and the resync just brings the persisted phase back in line with the unpaired state the device is genuinely in. The pairing flow re-requests a code on reconnect, so there is nothing to preserve across the reboot.

3. **Comment nit.** Fixed in 602b4dd9. `resetPicoHardware` now reads "in case the Pi or Pico rebooted mid-call" since it is invoked on both paths.

PR #746, opened 2026-06-11
feat(digitsd): drop the never-sent contacts/sync signaling messages

justinlindh:
## Why

The digitsd signaling protocol defines `sync`, `contacts`, and `contacts_updated` message types, and the daemon carries a full receive path for server-pushed contact lists: a `Contacts` field on the wire envelope, a `ContactEntry` type, a dispatch case that updates the cache and swaps the dial-filter checker, and the cache's `Update`/`Save` write half.

The server has never sent any of these. `git log -S '"contacts"' -- server` finds no commit that ever produced them, the server's `protocol.go` has no matching constants, and the server-side contacts tables were dropped in schema v6 when households/lines replaced them. The digitsd protocol header says `must match server/internal/signaling/protocol.go`; these three types are exactly the mismatch.

## What

- Remove `TypeSync`, `TypeContacts`, `TypeContactsUpdated`, `ContactEntry`, and the `Contacts` envelope field from `internal/signal/protocol.go`.
- Remove the contacts dispatch case, the `contactsCache` dispatcher field, and `SetContactChecker` from the dispatch-facing controller interface (main still calls it on the concrete controller).
- Trim `contacts.Cache` to its live surface: `Load`, `IsContact`, `Count`. The write half (`Update`/`Save`) existed only for the server push.
- Update tests accordingly; cache tests now exercise the load path directly, including a new invalid-JSON case.

## Why this scope

The dial-safelist feature itself stays: a hand-placed `contacts.json` is still loaded at startup and wired into the controller exactly as before, so device behavior is unchanged. Going smaller (constants only) would leave dead dispatch and cache code behind; going bigger (removing the safelist entirely) would delete a working, product-relevant feature rather than dead protocol surface.

Validated with `go build ./... && go test ./... && golangci-lint run ./...` in `pi/digitsd`.

### DIFF
--- a/pi/digitsd/cmd/digitsd/dispatch.go
+++ b/pi/digitsd/cmd/digitsd/dispatch.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/justinlindh/digits/pi/digitsd/internal/config"
-	"github.com/justinlindh/digits/pi/digitsd/internal/contacts"
 	"github.com/justinlindh/digits/pi/digitsd/internal/devmode"
 	"github.com/justinlindh/digits/pi/digitsd/internal/phone"
 	sigclient "github.com/justinlindh/digits/pi/digitsd/internal/signal"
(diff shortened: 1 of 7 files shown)


justinlindh:
@claude

PR #749, opened 2026-06-11
feat(server,digitsd,image): add web-app developer-mode toggle (SSH + dev UI)

justinlindh:
## Summary

Adds an admin-only "Developer mode" toggle on the line detail page that turns SSH and the on-device dev web UI on or off at runtime, no reflash required. Enabling prompts for an SSH password; disabling stops SSH and tears down the standing privilege it created.

- **server**: new `dev_mode` signaling command + `DevModePassword` field; admin-only `POST /phones/{number}/dev-mode` and owner-scope `GET /phones/{number}/dev-mode-status`; a "Developer mode" card in both the intercom and answering-machine themes (off: set an SSH password to enable; on: shows `ssh dev@<lan-ip>`, keeps the `:8080` admin-panel link, and a disable button). Card is gated to household admins.
- **digitsd**: handles the `dev_mode` command, applies it via a single privileged helper, manages the `:8080` dev-UI listener at runtime (no restart), and re-reports `device_info` so the UI reflects the new state. The dev-mode lifecycle is refactored into a small `devModeManager`.
- **image**: new `/usr/local/bin/digits-devmode` helper behind one sudoers allowlist entry. It owns all root work: SSH host keys, the `dev` user + password, `ssh.service`, and the password-auth drop-in. `enable` writes the on-flag last (only after `ssh.service` is confirmed active) inside a single rw window guarded by an EXIT trap; `disable` removes the passwordless-sudo and password-auth drop-ins and kills the dev user's live sessions, so "off" leaves no standing privilege. The helper is embedded into the digitsd binary (extracted at runtime on update); the sudoers entry ships with the next image build.

How it works end to end: admin submits the password form -> server sends `dev_mode` over the authenticated signaling channel (Redis fan-out reaches the device on any replica) -> digitsd runs `sudo digits-devmode enable`, starts `:8080`, re-reports `device_info` -> the UI polls `dev-mode-status` and reloads once the device confirms.

## Notes

- The SSH password crosses server->device over the already-authenticated WSS channel and, on the cross-replica path, transits the in-cluster Redis bridge (the same channel that already carries device tokens and pairing codes). It is passed to the helper on stdin, never argv.
- Release scopes: `server` (web toggle) and `digitsd` (handler + embedded helper) both need releases; `image` carries the sudoers entry, which ships with the next image build.

## Test plan

- [x] `make test-integration` + `golangci-lint` (server), `make test` + `golangci-lint` (digitsd), `shellcheck -S warning` (helper)
- [x] unit tests: `devModeManager` enable/disable/listener lifecycle, password validation, template parse
- [ ] Live on a real phone via the prod UI (post-deploy): enable -> SSH in as `dev` + `:8080` reachable; disable -> SSH refused + `:8080` gone; both themes. Screenshots to follow.

### DIFF
--- a/pi/digitsd/cmd/digitsd/devmode.go
+++ b/pi/digitsd/cmd/digitsd/devmode.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"embed"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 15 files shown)


justinlindh:
@claude

justinlindh:
Thanks for the review. Addressed in `5870a3d5`:

**1. `disable` ran `systemctl disable` on the read-only rootfs (medium):** confirmed and fixed. The runtime teardown (`systemctl stop` + `pkill -u dev`) now runs before the rw window, and the persistent `systemctl disable ssh.service` moved inside the `remount_rw`...`remount_ro` window next to the drop-in removal, mirroring `cmd_enable`'s `systemctl enable`. So the boot-time wants-symlink is actually removed and ssh.service no longer auto-starts after a disable+reboot.

**2. Concurrent `dev_mode` commands not serialized (minor):** fixed. `devModeManager.Enable`/`Disable` now hold `mu` across the `apply` (helper) call, not just the listener mutation, so two rapid toggles can't run two `digits-devmode` processes racing on the rootfs remount. The transition is now strictly serial.

**3. Helper byte-vs-rune floor (nit):** left as-is per your call. The helper's `${#password}` is only a backstop lower bound; the web handler's rune-based `validateDevModePassword` is the real gate. Not worth the asymmetry churn.

PR #751, opened 2026-06-11
fix(server): align developer-mode card with the device-controls group

justinlindh:
## Summary

Follow-up to #749 polishing the developer-mode card across all three themes, plus password-rule UX.

**Styling**
- The dev-mode block carried its own heading the sibling Recovery block does not have. In intercom (and dialup, which shares the same content block) it was a `field__label`; in the answering-machine theme it was an `am-plate__label`, which is a plate-level tab and so rendered as the heading for the entire Device controls plate.
- Intercom/dialup: drop the heading; self-labeling button first, muted hint beneath, matching the Recovery block.
- Answering-machine: move developer mode into its own `am-plate` (the correct use of `am-plate__label`), matching how that theme separates Voicemail, Voice style, and Actions into dedicated plates.

**Password UX**
- The SSH password input gets `minlength=8`, `maxlength=72`, `required`, and an inline error that appears on an invalid Enable attempt and clears on input, in both forms. The 8-to-72 hint stays as guidance; the server's `validateDevModePassword` remains the authority.

No behavior change beyond presentation/validation feedback.

## Test plan

- [x] `make test-integration` + `golangci-lint` (server), template-parse unit test
- [x] Visual on prod across intercom, dialup, and answering-machine

### DIFF
--- a/server/internal/web/templates/phone-detail.html
+++ b/server/internal/web/templates/phone-detail.html
@@ -393,7 +393,7 @@
             </div>
             <div id="devmode-enable-form" class="hidden">
               <label for="devmode-password" class="field__label" style="display: block; margin: 8px 0 4px;">SSH password</label>
-              <input type="password" id="devmode-password" class="field__input" autocomplete="new-password" minlength="8" maxlength="72" required aria-describedby="devmode-pw-hint" oninput="clearDevModePwError()" style="max-width: 280px;" placeholder="At least 8 characters">
+              <input type="password" id="devmode-password" class="field__input" autocomplete="new-password" maxlength="72" aria-describedby="devmode-pw-hint devmode-pw-error" oninput="clearDevModePwError()" style="max-width: 280px;" placeholder="At least 8 characters">
               <span id="devmode-pw-hint" class="muted" style="display: block; font-size: 11px; margin-top: 4px;">8 to 72 characters. Used to log in as <code>dev</code> over SSH.</span>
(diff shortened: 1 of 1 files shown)


justinlindh:
Before/after.

**Answering-machine theme, before.** The `am-plate__label` hijacked the whole Device controls plate, so the group heading read "DEVELOPER MODE":

am before

**Intercom theme, after.** The dev-mode block now matches the Recovery block: self-labeling button first, muted hint beneath, no stray sub-heading:

intercom after

justinlindh:
@claude

justinlindh:
Answering-machine theme, after the fix. Developer mode is now its own plate (consistent with Device controls and Actions), no longer hijacking the Device controls heading:

am after

Dialup ("Online service 1997") shares the intercom content block, so the same fix applies there:

dialup after

justinlindh:
@claude

justinlindh:
Thanks, addressed in `341d6f55`:

1. **Inert `required`/`minlength`:** removed both. You're right that with a `type=button` + `fetch` Enable there's no submit path, so native constraint validation never ran. The JS length check and the server's `validateDevModePassword` are the authority; `maxlength=72` stays since it's enforced as-you-type.
2. **`>72` branch unreachable:** noted, left as harmless defense-in-depth behind `maxlength`. The "8 to 72" copy still reads correctly (only the too-short case can fire from the UI).
3. **aria-describedby:** intercom input now references `devmode-pw-hint devmode-pw-error`; the AM input gained `aria-describedby="am-devmode-pw-error"` so the validation message is announced.
4. **Off-state copy divergence:** aligned the AM hint to the intercom wording ("You'll set an SSH password.").

PR #752, opened 2026-06-11
fix(image): disable ssh.service on production images

justinlindh:
## Summary

Follow-up to the developer-mode toggle (#749, #751). The runtime toggle assumes production images ship with SSH **off** and only turn it on when an admin enables dev mode. The build never actually asserted that: it explicitly disables/masks every other unwanted service (networkd, bluetooth, hostapd, avahi, polkit, dphys-swapfile, ...) but left `ssh.service` in whatever state the base image shipped.

- Explicitly `hostside_disable_service ssh.service` on the non-dev path, so a production device never runs sshd until dev mode is enabled at runtime. The `--dev` block (step 22) still re-enables it for dev images, and the `digits-devmode` helper enables it on demand for production devices (the helper already `systemctl unmask`s + `enable`s it).
- Correct a stale comment in `digits-first-boot.sh`: `/etc/ssh` is **not** a bind mount. The generated fstab only binds `/var/log`, `/tmp`, and `/home/digits`; dev images symlink `/etc/ssh/ssh_host_*` to `/data/ssh` instead. (This matches what's actually on-device.)

## Why not the larger consolidation

I'd floated folding the `--dev` build's dev-mode setup into the `digits-devmode` helper as a single source of truth. Investigating the boot path, that's riskier than it looked and not safely testable here:

- The SSH host-key mechanism is genuinely tangled: `fstab.fragment` documents an `/etc/ssh` bind mount that the generated fstab does not create, `digits-first-boot.sh` runs `ssh-keygen -A`, and the `--dev` block uses `/etc/ssh -> /data/ssh` symlinks. Untangling that touches first-boot and factory-reset behavior.
- Verifying it correctly requires flashing a dev image and exercising first boot + factory reset on real hardware, which isn't feasible in this change.

The current build/helper duplication is low-risk and working, so I kept this PR to the safe, high-value hardening rather than a boot-path refactor I can't fully verify.

## Test plan

- [x] `shellcheck -S warning -x` clean on `build-image.sh` and `digits-first-boot.sh`
- [x] Uses the same `hostside_disable_service` helper as the ~15 other disabled services
- [ ] On-hardware: build a production image and confirm sshd is not listening on first boot (not run here; no flash in this change)

### DIFF
--- a/pi/image/init-data.sh
+++ b/pi/image/init-data.sh
@@ -17,7 +17,7 @@
 #   ├── wifi/             # NetworkManager connection backups (digitsd-managed)
 #   ├── log/              # persistent logs (bind → /var/log)
 #   ├── tmp/              # tmp files (bind → /tmp)
-#   └── ssh/              # SSH host keys (bind → /etc/ssh)
+#   └── ssh/              # SSH host keys for developer mode (/etc/ssh/ssh_host_* symlink here)
 
 set -euo pipefail
 
@@ -77,7 +77,7 @@ install -d -m 2755 -o root -g adm "${MOUNT_POINT}/log/journal"
 # tmp/ — tmp files (bind-mounted to /tmp)
 install -d -m 1777 -o root -g root "${MOUNT_POINT}/tmp"
 
-# ssh/ — SSH host keys (bind-mounted to /etc/ssh)
+# ssh/ — SSH host keys for developer mode (/etc/ssh/ssh_host_* symlink here, not a bind mount; empty on production)
 install -d -m 755 -o root -g root "${MOUNT_POINT}/ssh"
 
 # ── create placeholder config.json ───────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


justinlindh:
@claude

justinlindh:
Thanks. Folded in the straggler fix in `690979ce`:

- `pi/image/rootfs-overlay/etc/fstab.fragment`: replaced the bogus `/data/ssh /etc/ssh` bind entry with a note that /etc/ssh is intentionally not bind-mounted.
- `pi/image/init-data.sh` (dir-tree comment + the `ssh/` section): corrected to "symlinked from /etc/ssh for dev mode, not a bind mount; empty on production."

So all three files now tell the same correct story.

Non-blocking notes acknowledged and skipped intentionally:
- **disable vs mask:** keeping `disable` (consistent with the other network services; the helper unmasks anyway).
- **ssh.socket / trixie:** YAGNI for now: the current base is Bookworm where sshd runs via `ssh.service`. If a future base bump moves to socket activation, the socket would need disabling too.

PR #765, opened 2026-06-19
fix(server): correct email template width and tidy server internals

justinlindh:
## What this is

A code cleanliness pass over the `server/` module: a review of idiomatic Go, dead/stale code, project layout, and test coverage, followed by fixes for the concrete findings. Includes one genuine user-visible fix (email rendering) alongside several internal refactors.

### Why 88 before

The module was already in strong shape: `go build`, `go vet`, `golangci-lint` (standard) and the full test suite all pass clean, `deadcode` is clean, there are no TODO/FIXME markers, the layout is standard (`cmd/` thin mains over `internal/` packages, embedded templates and static assets), and the code is idiomatic throughout (errors wrapped with `%w`, `errors.Is`/`errors.As` against sentinels, rows closed with `rows.Err()` checked, no naked returns in long functions, no else-after-return chains). Test coverage is broad across unit, integration (build-tagged), and Playwright e2e tiers.

The deductions were for a handful of real issues:

- A conference summary `SELECT` plus its scan-into-struct was copy-pasted across two files, with the column list and scan that must stay in lockstep being exactly the duplicated parts (drift risk).
- The find-or-create-user sequence, including the security-sensitive distinction between a missing user and a transient DB error, was hand-copied across two handlers.
- Em dashes in source comments and a log string, against the project's own hard no-em-dashes rule.
- Dead code: never-populated `Release.Commit` / `Release.ReleasedAt` fields, an always-true `if span != nil` guard, a no-value `splitFields` wrapper.
- Stale comments: an "All four" rate-limiter count (there are six) and an `autodeploy` package doc describing an arm64 binary download and process hot-swap the package does not do.
- A broken `width:100%.` declaration in the email template (the stray period makes clients drop the rule, losing responsive width).

### What changed (and why 93 after)

| Commit | Change |
| --- | --- |
| `refactor` | Consolidate the duplicated conference summary query and scan into `conferenceSummaryColumns`/`conferenceSummaryGroupBy` consts and a `scanConferenceSummary` helper, mirroring the existing `callColumns`/`scanCallRows` pattern. |
| `refactor` | Move find-or-create-user into `Store.GetOrCreateUserByEmail`, which also reports whether a user was created so dev-session can mark the theme chosen. The Google callback keeps its own three-way lookup (it links by Google ID and creates with the OAuth name). |
| `refactor` | Make `buildChangelogSection` a `*Handler` method instead of threading `*Handler` as a trailing argument, matching its peers. |
| `fix` | Remove the stray period from the email table `width:100%` so the responsive width applies on narrow screens. |
| `refactor` | Remove dead code (dead nil guard, unused `Release` fields and the `time` import, the `splitFields` wrapper) and fix stale comments and em dashes in source. |
| `refactor` | Use the shared `dbutil.RowScanner` for the conference scan instead of a local redeclaration (found by the `/simplify` pass). |

After the pass: `build`, `vet`, `golangci-lint`, and the full test suite still pass clean, the duplication and drift risk are gone, production Go source is em-dash-clean, and the stale docs match the code.

### Deliberately not in scope

- `/healthz` on the metrics listener still inlines `{"status":"ok"}` rather than reusing `httputil.Healthz`: switching it would change that ops endpoint's response body (the shared helper adds a version field), a behavior change to a probe target.
- `events.Broadcaster.Notify` / `notifyLocal` share a fan-out loop: left as is, since deduping it touches lock ordering for a roughly six-line saving.
- `household.LinkStore` returns ad-hoc `errors.New` for not-found and conflict cases while the sibling `InviteStore` uses sentinels: no caller distinguishes them via `errors.Is` today (they are rendered as text), so adding sentinels now would be speculative API surface (YAGNI).
- Em dashes remain in test files and in CSS/HTML comments (and the dev-only test client): these are comment and dev-tooling punctuation with no Go-code-quality or functional signal, and template copy risks e2e assertions, so they are out of scope for this production-source pass.

## Testing

- `go build ./...`, `go vet ./...`: clean.
- `golangci-lint run ./...` (standard linters): 0 issues.
- `go test ./...`: all packages pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YU4M1N81pWAzpPzWDP3h35)_

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -227,7 +227,7 @@ func run(ctx context.Context) error {
 		return fmt.Errorf("parse login template: %w", err)
 	}
 	if cfg.DevMode {
-		slog.Warn("dev mode enabled — magic link URLs will be logged to stdout")
+		slog.Warn("dev mode enabled: magic link URLs will be logged to stdout")
 	}
 	authHandlers := auth.NewHandlers(authStore, googleAuth, emailSender, cfg.BaseURL, cfg.CookieDomain, loginTmpl, cfg.DevMode)
 
--- a/server/internal/auth/handlers.go
+++ b/server/internal/auth/handlers.go
@@ -1,7 +1,6 @@
 package auth
 
 import (
-	"errors"
 	"fmt"
 	"html/template"
 	"log/slog"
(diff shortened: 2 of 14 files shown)


justinlindh:
@claude

PR #779, opened 2026-06-23
feat(server): toggle between devices joined to a line

justinlindh:
## Why

When two devices are joined to one line, the line detail page only ever showed **one** of them. The operator panel (versions, status, LAN IP, dev/admin link, restart/recovery/dev-mode controls) rendered an arbitrary device (`allInfos[0]`), so a secondary handset like `Justin Office` on `245-6390` was unreachable: no way to see its software version, its dev link, or to restart just that handset. Worse, the operator actions fanned out over `SendTo(number)`, so "Reboot device" silently rebooted **every** handset on the line.

## What

A segmented device toggle on the operator column lets you switch between the handsets joined to a line and view plus control each one individually.

- **Left column stays shared** (line settings: voice style, ringer, quiet hours, voicemail). These are line-level and apply to every handset.
- **Right column is per-device.** A toggle (one segment per handset, shown only when a line has 2+ devices) htmx-swaps an operator-panel partial scoped to the selected device.
- **Commands target the selected handset.** Restart, reboot, recovery, developer mode, Pi/firmware update, and ring test now route to that device's `HardwareID` via the hub's `SendToHardware`, with the `hardware_id` validated to belong to the line. This also fixes the prior behavior where those actions hit every joined handset.
- **Per-device status.** Update progress and online state are keyed by hardware ID, and last-seen is sourced from the selected device's own row (not the line-level aggregate), so a swap matches a full reload.
- Single-device lines are unchanged: no toggle, same behavior.
- All three themes get the toggle.

## Screenshots

Each pair is the same line with two handsets reporting distinct versions: selecting **Kitchen** (Pi 1.21.0 / fw 0.9.0 / 192.168.2.142) vs **Hallway** (Pi 1.20.0 / fw 0.8.1 / 192.168.2.150) swaps the operator panel to that handset while the shared line settings stay put.

### Intercom (default)

| Kitchen selected | Hallway selected |
|---|---|
| intercom kitchen | intercom hallway |

### Dialup

| Kitchen selected | Hallway selected |
|---|---|
| dialup kitchen | dialup hallway |

### Answering machine

| Kitchen selected | Hallway selected |
|---|---|
| am kitchen | am hallway |

## Testing

- Unit suite green; integration suite green (`-race -p 1`, CI-style).
- New coverage: `buildDeviceViews` selection logic, per-device command routing and foreign-`hardware_id` rejection, operator-panel swap returns the requested device's data without leaking a sibling's (both themes), per-hardware update status.
- Verified end to end against a local stack with two live handsets reporting distinct versions (the screenshots above).

### DIFF
--- a/server/internal/signaling/hub.go
+++ b/server/internal/signaling/hub.go
@@ -74,7 +74,7 @@ type Hub struct {
 	mu           sync.RWMutex
 	conns        map[string][]*Conn               // phone number -> connections (multiple devices per line)
 	hwConns      map[string]*Conn                 // hardware ID -> connection
-	updateStatus map[string]*UpdateStatusSnapshot // phone number -> last update status
+	updateStatus map[string]*UpdateStatusSnapshot // hardware id -> last update status
 	// voicemailUnheard tracks the per-handset unheard-voicemail count last
 	// reported by each device. Outer key is the phone number; inner key is
 	// the originating handset's hardware ID. Per-handset because voicemail
@@ -467,10 +467,6 @@ func (h *Hub) RekeyNumber(oldNumber, newNumber string) {
 		h.conns[newNumber] = cs
 		delete(h.conns, oldNumber)
 	}
-	if us, ok := h.updateStatus[oldNumber]; ok {
-		h.updateStatus[newNumber] = us
-		delete(h.updateStatus, oldNumber)
-	}
 	if vm, ok := h.voicemailUnheard[oldNumber]; ok {
 		h.voicemailUnheard[newNumber] = vm
 		delete(h.voicemailUnheard, oldNumber)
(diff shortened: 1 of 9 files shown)


justinlindh:
@claude

justinlindh:
Thanks, addressed:

**1. `IsHardwareOnline` pod-local (multi-pod).** Fixed in a39e030c. Multi-pod is a real deployment target, so `IsHardwareOnline` now delegates to a Redis-backed presence check (`EXISTS` on the per-device key `AllDeviceInfo` already reads) when a shared device-state store is configured, falling back to the local connection map otherwise. Added a miniredis test covering cross-pod presence and that one handset going offline doesn't affect its sibling.

**2. Header badge initial vs poll.** Resolved by #1: the selected device's `.Online` is now cross-pod, so it agrees with the line-level poll.

**3. Legacy no-row connection loses last-seen.** Fixed in a39e030c: the empty-id fallback in `buildOperatorData` now uses the line-level `hub.LastSeenAt` again.

**4. JSON error from the htmx GET.** Skipping. `requireLineDevice`'s JSON 400 is the right shape for its primary callers (the `fetch()`-based action/status handlers that read `data.error`); the operator GET only reaches that branch on a hand-tampered `hardware_id`, since the toggle only ever emits ids that belong to the line. Not worth a separate validation path.

Separately, the answering-machine toggle was rendering full-width (the `inline-flex` control stretched to fill its flex-column plate); fixed in 237011aa + 4969c9fe with `width: fit-content` (verified 155px). PR screenshots updated.

PR #781, opened 2026-06-23
feat(server): add ICE/SDP relay observability to signaling

justinlindh:
## What

Makes call media negotiation debuggable end to end. signald relays every SDP and ICE message between the two phones, but an ICE candidate previously passed through opaquely: an operator could not tell whether phones were exchanging host, srflx, or relay candidates, what transport they used, or which joined device on a multi-device line was involved. The server is the one place both ends' candidates converge, so it is the right place to make this visible.

## Why

The hardest media failures were invisible on the server:

- "Call connects but there is no audio" gives no server signal about which candidate types were even exchanged.
- Silent TURN fallback (phones giving up on host/srflx and relaying everything) looked identical to a healthy direct call.
- On a shared line (multiple devices on one number) the call-lifecycle logs only carried the line number, so you could not tell which handset initiated or answered.
- ICE-server issuance was completely unlogged, so a pod handing out STUN-only because of a TURN misconfiguration was undetectable from logs or metrics.

## What changed

- **Candidate parser + SDP summarizer** (`internal/signaling/candidate.go`, unit-tested). `ParseCandidate` extracts type, transport, address:port, and raddr/rport from the raw SDP candidate attribute the phone trickles (the value of `ICECandidate.ToJSON().Candidate`). `SummarizeSDP` counts m-lines and embedded candidates without retaining the SDP body. Both are lenient: malformed or end-of-candidates input yields a zero value, never a panic.
- **Per-candidate Debug logging** in the relay forward path with the parsed candidate fields plus `from`/`to`/`hardware_id`/`call_id`. Debug keeps high-volume detail off the Info stream but one filter away.
- **Info SDP relay logging** on both the offer and answer paths, carrying the same attribution plus a content-free shape summary (`m_lines`, `embedded_candidates`, `sdp_bytes`). The SDP body is never logged.
- **Info ICE-server issuance logging** in `handleRequestICE`, keyed by number + `hardware_id`, recording whether STUN and TURN were handed out and the TURN transports offered. The TURN username and credential are never logged.
- **hardware_id on call-initiated and call-answered logs** so a call on a shared line is attributable to the specific handset.
- **Two Prometheus counters**: `ice_candidates_relayed_total` (by candidate type and transport) and `ice_servers_issued_total` (by whether TURN was included). Both use closed label sets that collapse unknown values to `other`, matching the existing `signaling_errors_total` contract, so a malformed candidate parsed from untrusted device input can never widen the label space. Candidate address and port are never labels.

## Privacy

No TURN credentials, device tokens, or SDP bodies are logged or recorded as metrics. Candidate address:port appears only in slog Debug, consistent with existing `local_addr` logging, and never in trace span attributes, preserving the tracing privacy contract in `internal/tracing`.

## Tests

- New parser table tests (host/srflx/relay/prflx, IPv6, TCP, malformed, odd trailing tokens).
- New relay tests asserting the candidate metric fires with the parsed type/transport, that an empty (end-of-candidates) candidate is not counted, and that the candidate is still relayed verbatim.
- New metrics tests for both counters including the unknown-collapses-to-other path.
- `make test` green; `golangci-lint run ./...` (v2.1.6) reports 0 issues.

### DIFF
--- a/server/cmd/signald/main.go
+++ b/server/cmd/signald/main.go
@@ -185,6 +185,7 @@ func run(ctx context.Context) error {
 	relay := signaling.NewRelay(hub, tracker, line.NewAuthorizer(database), signaling.NewLineStoreAdapter(lineStore))
 	relay.HealthStore = healthStore
 	relay.Errors = mreg
+	relay.Metrics = mreg
 	tracker.SetCallEndObserver(relay)
 	hub.SetReconnectHook(relay.HandleRemoteReconnect)
 	hub.SetDropHook(func() { mreg.ObserveSignalingError("send_buffer_full") })
--- a/server/internal/metrics/metrics.go
+++ b/server/internal/metrics/metrics.go
@@ -14,6 +14,11 @@
 //   - Active calls (a gauge of in-flight calls, count only).
 //   - Signaling errors by category (e.g. turn_alloc_failed, ice_timeout).
 //     The category is a fixed enum; no peer identity, number, or content.
+//   - ICE candidates relayed, labeled by candidate type (host/srflx/prflx/
+//     relay) and transport (udp/tcp). Both labels are fixed enums derived
+//     from the parsed candidate; no address, port, or peer identity.
+//   - ICE-server responses issued, labeled only by whether TURN was included.
+//     No device identity and never the TURN username or credential.
 //   - Build info as a static gauge labeled with the version and short commit.
 //   - Go runtime and process collectors (goroutines, GC pauses, memory, fd
 //     count) provided by promhttp / collectors. Same data the Go runtime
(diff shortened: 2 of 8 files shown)


justinlindh:
@claude

justinlindh:
Thanks. Addressed the two actionable observations:

**1. `sdp_kind` asymmetry.** Normalized: an offer now logs `sdp_kind="offer"` instead of the wire type `sdp`, symmetric with `answer`.

**2. Malformed candidates invisible in metrics.** `ObserveICECandidate` now fires for every non-empty candidate. An unparseable one carries empty type/transport, which the metric's label allowlist collapses to `other/other`, so a rising malformed-candidate rate shows on a dashboard, not only in Debug logs. Added `TestRelayObservesMalformedCandidateAsOther`.

**3. raddr/rport pairing invariant.** Left as-is per your note: RFC 8839 extension attributes are always name-value pairs, and `TestParseCandidateOddTrailingTokens` covers the lone-trailing-token case. No behavior change.

PR #785, opened 2026-06-24
fix(server): fan out line messages to devices on every pod

justinlindh:
## What

Line-targeted Hub sends now reach every device on the line regardless of which signald pod terminates each device's WebSocket.

## Root cause

`Hub.SendTo` (the fan-out path behind ring, offer/answer, hangup, busy, conference, call-return, quiet-hours) delivered only to connections on the pod that handled the message, and skipped the Redis cross-pod publish whenever at least one device for the line was connected locally:

```go
conns := h.conns[number]      // local conns on THIS pod only
if len(conns) == 0 {
    bridge.Publish(...)       // cross-pod ONLY when zero local conns
    return nil
}
for _, conn := range conns { conn.Send <- data }  // local only, no publish
return nil
```

signald runs 3 replicas, so a line's devices are spread across pods. An incoming call to a multi-device line rang only the device(s) that happened to share a pod with the caller's relay; siblings on other pods were silently dropped. This was observed in prod: a line with three phones rang only one (then two on retry, depending on pod placement). `ring_test` always worked because it uses `SendToHardware`, which falls back to Redis when the device is not local.

## Fix

`SendTo` now mirrors `Broadcast`: it always delivers to local connections AND publishes to Redis. The Redis subscriber already filters out the origin pod's own envelope (by `PodID`), so local devices are not double-sent, and remote pods fan out to their local connections via `deliverFromRedis`.

`SendToWithTimeout` (reliable ICE-restart delivery) had the identical gap: a sibling device connected locally suppressed the cross-pod publish, so an in-call peer on another pod could miss an ICE restart. It now publishes once up front and still retries local buffer-full delivery against the deadline.

## Tests

- Rewrote `TestSendToLocalFastPathSkipsRedis` (which asserted the buggy behavior) as `TestSendToAlsoPublishesToRedisWithLocalConn`: a local conn receives the message AND a cross-pod envelope is published.
- Added `TestSendToWithTimeoutPublishesToRedisWithLocalConn`.
- Full server unit suite passes; `golangci-lint` v2.11 clean on the package.

### DIFF
--- a/server/internal/signaling/hub.go
+++ b/server/internal/signaling/hub.go
@@ -515,7 +515,11 @@ var ErrSendTimeout = errors.New("send timed out: buffer full")
 // phones on the line. A line's devices can be spread across pods, so SendTo
 // always both delivers to local connections AND, when Redis is configured,
 // publishes for cross-pod delivery. The Redis subscriber filters out our own
-// pod's envelope, so local devices are never double-sent.
+// pod's envelope, so local devices are never double-sent. Publishing is
+// unconditional rather than gated on per-line presence: a presence read per
+// send would add a TOCTOU window (a device can connect on another pod between
+// the check and the send) to shave a single missed map lookup off an already
+// cheap call-signaling path, so the gate is deliberately omitted.
 // Returns ErrNotConnected only when no devices are connected locally AND
 // Redis is not configured.
 func (h *Hub) SendTo(number string, msg *Message) error {
--- a/server/internal/signaling/hub.go
+++ b/server/internal/signaling/hub.go
@@ -509,10 +509,15 @@ func (h *Hub) ConnectionCount(number string) int {
 // buffer does not drain within the deadline.
 var ErrSendTimeout = errors.New("send timed out: buffer full")
 
-// SendTo marshals msg and sends it to every device on the given line number.
-// This is the POTS extension model: a ring reaches all phones on the line.
+// SendTo marshals msg and sends it to every device on the given line number,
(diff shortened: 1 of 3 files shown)


justinlindh:
@claude please review this fix. Focus on the cross-pod fan-out correctness: double-delivery safety (origin-pod self-filter by PodID), and whether always-publishing SendTo to Redis has any ordering or load concerns given it now carries the full line-targeted signaling surface.

justinlindh:
Thanks, addressed the one actionable note. Pushed `5388df49` documenting on `SendTo` why the publish is unconditional rather than presence-gated (the TOCTOU window vs. a single missed map lookup tradeoff you flagged), so the next reader doesn't re-ask.

On the rest, all confirmed and intentional:
- **Contract change** (`SendTo` no longer returns `ErrNotConnected` with Redis on): correct, and the one caller that inspects it (`handler_phones.go`) was already only reachable when Redis is unconfigured. Captured in the doc comments.
- **Publish-before vs publish-after asymmetry**: left as-is. `SendToWithTimeout` publishing up front is the cleaner spot (keeps cross-pod ICE-restart latency out from behind the local retry deadline), and the up-front `RLock` for `bridge` is what makes "publish exactly once" clean.
- **Unique pod IDs now load-bearing for all line signaling**: noted. Pod identity is unique per replica in prod; nothing in this PR changes that assumption.
- **`ErrSendTimeout` over-reporting when local buffers stay full but remote got it**: pre-existing local-delivery semantics, leaving it.

PR #786, opened 2026-06-24
fix(server): self-heal device line-number drift instead of reconciling forever

justinlindh:
## Root cause

Devices store their line number locally and claim it on every WebSocket register. When a line is renumbered, moved, or rejoined, the device keeps claiming its OLD number. `handler_ws.go` already "reconciles" this on the fly: it looks up the device's `BoundLineNumber`, and when the claimed `Number` differs it logs `ws register reconciled to bound line` and registers the device under the bound number instead.

The problem is the reconcile is a perpetual band-aid. The device never learns its correct number, so it re-claims the stale one on the next register, and the server re-reconciles, forever. Real example in prod: device `df994964-ab73-4581-aa01-c16376f32a59` ("Justin Office") claims `2451234` but is bound to `2456390`, reconciling on every register.

## Fix

Make the drift self-heal. When the server reconciles, it pushes the corrected number back to the device so `digitsd` persists it and stops claiming the stale one.

- New `line_renumber` message type (Server -> Phone), carrying the bound number in `Number`. Added to both protocol definitions.
- `server/internal/web/handler_ws.go`: after the reconciled connection is registered, enqueue a single `line_renumber` with the bound number on the connection's buffered send channel. It fires only on an actual reconcile; a correctly-numbered register stays silent.
- `pi/digitsd`: a `line_renumber` handler writes the authoritative number to config (`PhoneNumber`, persisted to `/data/digits/config.json` via the existing atomic Save), then restarts so the daemon re-registers under it. The handler is a no-op when the incoming number is empty or already matches the stored one, so a redundant or echoed message can never cause a restart loop. The restart exit is indirected through `reexecProcess` so the path is unit-testable.

I considered reusing the existing `paired` message (it already persists `Number`), but its `digitsd` handler is gated on a non-empty `DeviceToken` and unconditionally fires `StateSet("PAIRED")` plus a dial tone. Replaying the pairing UX (audible dial tone on an idle, on-hook, already-paired phone) during a routine reconnect is wrong, so a narrow dedicated message keeps `paired` semantics clean.

I did not touch `hub.go` `SendTo`: the message is delivered directly on the connection's own send channel in `handleWS`.

## Tested

- `pi/digitsd`: new unit tests in `dispatch_test.go` prove a `line_renumber` with a different number persists to config (reloaded from disk) and requests a restart, and that a same/empty number is a no-op (no config write, no restart). Full `pi/digitsd` unit suite passes; `golangci-lint` clean.
- `server`: updated the existing reconcile integration test to assert the device receives exactly one `line_renumber` carrying the bound number, and added a test that a correctly-numbered register receives no renumber. WS and signaling integration tests pass against Postgres; full unit suite passes; `golangci-lint` clean.

### DIFF
--- a/pi/digitsd/cmd/digitsd/dispatch.go
+++ b/pi/digitsd/cmd/digitsd/dispatch.go
@@ -383,12 +383,18 @@ func (d *daemonCallbacks) handleSignal(msg *sigclient.Message) {
 			break
 		}
 		slog.Info("signal: line renumber", "old", d.cfg.PhoneNumber, "new", msg.Number)
+		// Persist to disk before committing the change in memory. If Save fails
+		// we revert so in-memory state never diverges from disk: the daemon
+		// stays on its old number (the server reconciles again next connect)
+		// rather than running half-applied until the next reload.
+		oldNumber := d.cfg.PhoneNumber
 		d.cfg.PhoneNumber = msg.Number
-		d.number = msg.Number
 		if err := d.cfg.Save(); err != nil {
+			d.cfg.PhoneNumber = oldNumber
 			slog.Warn("signal: line renumber -- failed to save config", "error", err)
 			break
 		}
+		d.number = msg.Number
 		slog.Info("signal: restarting to re-register on corrected line", "number", msg.Number, "config", d.cfg.Path())
 		go func() {
 			time.Sleep(1 * time.Second)
--- a/pi/digitsd/cmd/digitsd/dispatch.go
+++ b/pi/digitsd/cmd/digitsd/dispatch.go
@@ -372,6 +372,29 @@ func (d *daemonCallbacks) handleSignal(msg *sigclient.Message) {
 			}()
 		}
 
+	case sigclient.TypeLineRenumber:
+		// The server reconciled our register to a different bound line: our
+		// stored number is stale (line moved, joined, or renumbered). Persist
+		// the authoritative number and restart to re-register under it. Without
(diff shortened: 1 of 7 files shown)


justinlindh:
@claude please review. This self-heals device line-number drift: after the server reconciles a stale-numbered register to the bound line, it pushes a new `line_renumber` message and digitsd persists the corrected number and restarts to re-register. Focus on: (1) is restarting digitsd the right call vs an in-place re-register, (2) any loop risk if persistence or re-register fails, (3) the `conn.Send <-` enqueue happening before the write pump starts.

justinlindh:
Both addressed.

- **(3) blocking send (the merge-blocker):** moved the `line_renumber` enqueue to *before* `h.hub.Register` (`bd579029`... `handler_ws.go`). The conn isn't hub-visible yet, so nothing can contend for the fresh cap-32 buffer and the send always has room. Picked this over a non-blocking `select`/`default` because dropping a self-heal message would silently strand the device on its stale number. Zero behavior change otherwise.
- **(2) save ordering:** `dispatch.go` now sets the config field, saves, and only assigns `d.number` on success, reverting the config field if `Save` fails. On failure the daemon stays fully on its old number rather than half-applied. Logged old/new captured before the mutation.

You were right that restart-to-re-register relies on `Restart=always` (same as the `paired` and `restart:service` arms); no new dependency. The oscillating-bound-number pathology is server-side DB consistency, out of scope. Both modules build, test, and lint clean.

PR #800, opened 2026-06-29
fix(server): stop counting unreachable dials as signaling errors

justinlindh:
## What

The `DigitsSignalingErrors` critical alert paged. Root cause: a registered phone dialed a run of numbers that don't exist, then trickled SDP/ICE for those never-established calls. The relay recorded each as a signaling error (`peer_unreachable` for the call, `invalid_message` for the trailing SDP/ICE), and the alert fires on any nonzero error rate, so one phone dialing dead numbers paged on-call critical with nothing actually wrong.

Dialing an offline or unregistered peer is normal user behavior, not a server fault. This stops counting it as one.

## Changes

- `handleCall`: dialing an offline/unregistered number no longer emits `peer_unreachable`. The caller still gets the `phone not connected` reply.
- `handleDTMF`, `handleICERestart`, `handleAnswer`, `handleSignalingForward`: the "without active call" branches no longer emit `invalid_message`. These are benign races (candidates trickling past teardown) or the tail of an unreachable dial. The `ice_restart` path still sends its `no active call` reply.
- These stray-message logs drop from `Warn` to `Debug` since they're expected, not anomalies. (One phone produced ~80 `Warn` lines in 10 minutes.)
- `peer_unreachable` is now unemitted, so it's removed from the metric category allowlist.
- `docs/metrics.md` updated to describe the category set as genuine faults only.

`signaling_errors_total` keeps counting real faults: `auth_failed`, `call_setup_failed`, `relay_delivery`, malformed messages (empty destination), and the reserved TURN/ICE categories.

## Out of scope

The `DigitsSignalingErrors` alert rule itself (the `rate(...) > 0` threshold) lives in the deployment's alerting config, not this repo. This PR fixes what feeds the metric; the threshold can be revisited separately if a single genuine fault paging critical is still too sensitive.

## Test

- Replaced the two tests that asserted the old behavior (`TestRelayObservesPeerUnreachableOnOfflineCall`, `TestRelayObservesInvalidMessageOnICERestartWithoutCall`) with tests asserting no error is recorded while the caller-facing replies are preserved, covering call/dtmf/ice_restart/answer/sdp/ice.
- Full server unit suite passes; `go vet` and `golangci-lint run ./...` clean.

### DIFF
--- a/server/docs/metrics.md
+++ b/server/docs/metrics.md
@@ -50,13 +50,15 @@ All metric names are prefixed with `digits_signald_`.
 ### Signaling errors (signald only)
 
 - `digits_signald_signaling_errors_total{category}` (counter): signaling
-  errors observed by the relay. `category` is a closed set defined in code
+  errors observed by the relay. The set counts genuine server-side faults
+  only: normal user behavior, such as dialing a phone that is offline or
+  unregistered (and the SDP/ICE/DTMF that trails such a dial), is not an
+  error and is not counted. `category` is a closed set defined in code
   (see `internal/metrics/metrics.go`):
-  - `peer_unreachable`: a CALL was sent to a phone not connected.
   - `auth_failed`: the call authorizer denied the call.
   - `call_setup_failed`: the tracker failed to record a call initiation.
-  - `invalid_message`: the relay received a malformed or out-of-context
-    message (e.g. `ICE_RESTART` with no active call).
+  - `invalid_message`: the relay received a malformed message (e.g. one with
+    no destination).
   - `relay_delivery`: a forward to a peer's WebSocket failed.
   - `turn_alloc_failed`, `ice_timeout`: reserved for future use.
   - Any value not in the closed set collapses to `other`. The relay
--- a/server/internal/metrics/metrics.go
+++ b/server/internal/metrics/metrics.go
@@ -186,7 +186,6 @@ func (r *Registry) RegisterCallsGauge(read func() float64) {
 var validErrorCategories = map[string]struct{}{
 	"turn_alloc_failed": {},
 	"ice_timeout":       {},
-	"peer_unreachable":  {},
 	"call_setup_failed": {},
 	"auth_failed":       {},
 	"invalid_message":   {},
--- a/server/internal/signaling/relay.go
+++ b/server/internal/signaling/relay.go
@@ -556,7 +556,7 @@ func (r *Relay) handleSignalingForward(ctx context.Context, from string, msg *Me
 		// SDP/ICE for a call that isn't active: the peer was never reachable
 		// (dialed an offline or unregistered number) or the call already ended
 		// and candidates are still trickling in. Both are normal, not errors.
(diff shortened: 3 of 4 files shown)


justinlindh:
@claude

justinlindh:
Addressed the review.

**Item 1 (empty-destination `invalid_message` shadowed): fixed in `d64dd6ad`.** You were right: with a live tracker, `inCallOrConference(from, "")` returns false, so the without-active-call guard caught empty-`To` messages and dropped them as benign Debug, and the dedicated empty-`To` `invalid_message` branches only fired under a nil tracker (tests). Rather than tweak the doc, I moved the destination check ahead of the active-call guard: a new `rejectNoDest` guard at the top of `handleDTMF`/`handleICERestart`/`handleAnswer`/`handleSignalingForward` counts an empty-`To` message as `invalid_message` before the guard runs. That keeps the two cases distinct: empty destination is a malformed message (real fault, counted), a valid dial to a peer with no active call is normal (not counted). The redundant empty-`To` checks in `handleICERestart` and `forward` are gone. The `metrics.md` "no destination" example is accurate again, and a new test (`TestRelayObservesInvalidMessageOnEmptyDestination`) locks it in across all relay types.

**Nit (ice_restart reply not covered): fixed in `d64dd6ad`.** `TestRelayDoesNotErrorOnSignalingWithoutCall` now asserts the `no active call` reply is delivered for ice_restart (and that the other types send no reply).

**Nit (ANSWER in doc wording): skipped.** The `metrics.md` sentence describes what trails an outbound dial to a dead number, which is SDP/ICE/DTMF; an `answer` is the callee-side race (answering a call that just ended), a different scenario, so I left it out of that specific list intentionally rather than overstate it.

justinlindh:
Addressed in d64dd6ad.

**Item 1 (empty-destination `invalid_message` shadowed by the active-call guard):** fixed in code rather than docs. An empty destination is a genuinely malformed message (a real fault), distinct from a valid dial to a peer that isn't in a call (normal behavior), which is exactly the distinction this PR is about. Added a `rejectNoDest` guard that runs at the top of `handleDTMF`, `handleICERestart`, `handleAnswer`, and `handleSignalingForward`, before the `inCallOrConference` check, so an empty `To` is counted as `invalid_message` instead of collapsing into the guard's benign Debug drop. Removed the now-redundant empty-`To` checks in `handleICERestart` and `forward` so there is a single emit site. The `docs/metrics.md` "no destination" example is accurate again because it now actually fires in production.

**Nit (ice_restart reply not covered):** `TestRelayDoesNotErrorOnSignalingWithoutCall` now asserts the client-facing reply per type, so the `ice_restart` "no active call" reply is locked in alongside the no-error assertion.

**New test:** `TestRelayObservesInvalidMessageOnEmptyDestination` covers the empty-`To` path across all five relay types.

**Doc-wording nit (ANSWER):** left as-is. The sentence describes what trails an outbound dial to a dead number (SDP/ICE/DTMF); `answer` is a different scenario (a callee answering a call that just ended), so folding it into that phrase would be less accurate, not more.

Full server unit + integration suites pass; `go vet` and `golangci-lint run ./...` clean.

PR #809, opened 2026-07-02
fix(server): stop a same-hardware reconnect from dropping a live call

justinlindh:
## What

Fixes a race where a device reconnecting with the same `hardware_id` during an active 2-party call has that call torn down ~20 seconds later.

## Why

When a device reconnects while its previous websocket is still registered, `Hub.Register` replaces the old conn in place and the new conn runs `OnReconnect` before the old conn's read loop unwinds. The reconnect's `cancelGraceLocal` finds no timer yet (nothing to cancel). Then the old loop's `OnDisconnect` runs, sees the new sibling as the sole remaining connection (`ConnectionCount == 1`, so the `> 1` guard does not fire), finds a peer, and calls `startGraceTimer`. Nothing ever cancels that timer, so when `GraceWindow` (20s) expires it runs `endActiveCallsAsHangup` and drops the reconnected, still-active call.

This is common on mobile and flaky Wi-Fi, where a new socket lands before the server notices the stale one via the 45s pong deadline. The user sees a call drop about 20 seconds after a brief blip.

The fix routes disconnect through a new `Relay.OnConnClosed`, which skips teardown when the departing conn is no longer the hub's current connection for its line (identity check via the new `Hub.ConnIsCurrent`). A conn displaced by a same-hardware reconnect is not current, so its stale disconnect no longer starts an orphaned timer. `OnDisconnect` is unchanged and still holds the call open for genuine last-device drops.

## Test plan

- `go test ./internal/signaling/...` passes, including the existing grace-window suite.
- New `TestConnIsCurrent` covers the hub identity check across register, same-hardware replace, and unregister.
- New `TestOnConnClosedSkipsTeardownForSupersededReconnect` reproduces the race and asserts no orphaned grace timer is left and the call is not ended. Verified it fails without the guard (`an orphaned grace timer was started for a reconnected device`) and passes with it.
- New `TestOnConnClosedHoldsCallForGenuineDisconnect` guards against over-suppression: a real last-device disconnect still starts the grace timer.
- `go vet ./internal/signaling/... ./internal/web/...` clean; full server build passes.

---

### DIFF
--- a/server/internal/signaling/grace_test.go
+++ b/server/internal/signaling/grace_test.go
@@ -124,6 +124,66 @@ func TestOnDisconnectHoldsInCall2Party(t *testing.T) {
 	}
 }
 
+// A device that reconnects with the same hardware_id while its previous
+// connection is still registered must not have its live call torn down. The
+// new conn replaces the old in place and runs OnReconnect (which finds no
+// timer yet); then the old conn's read loop unwinds. OnConnClosed must notice
+// the old conn was superseded and skip teardown, so no orphaned grace timer is
+// left to expire and hang up the reconnected call.
+func TestOnConnClosedSkipsTeardownForSupersededReconnect(t *testing.T) {
+	hub := NewHub()
+	tracker := newMockTracker()
+	tracker.peers = map[string]string{"3140001": "3140002"}
+	relay := NewRelay(hub, tracker, nil, nil)
+	relay.GraceWindow = time.Hour // a wrongly-started timer would not fire during the test
+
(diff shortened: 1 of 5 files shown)


justinlindh:
@claude

justinlindh:
Good catch on the orphaned doc comment. Fixed in 10e1922: moved the `OnConnClosed` comment directly above its function and restored the `OnDisconnect` comment above `OnDisconnect`, so godoc attaches each to the right function.

---

PR #810, opened 2026-07-02
fix(digitsd): close pipeline output channel on stop to stop goroutine leak

justinlindh:
## What

Closes the audio pipeline's output channel in `Pipeline.Stop()` so the per-call encode/record goroutines exit instead of leaking.

## Why

`Pipeline.Stop()` did `close(p.stop); p.wg.Wait()` and closed the ALSA capture and denoiser, but never closed `p.outPCM`. Consumers iterate with `for frame := range OutFrames()` (the WebRTC encode loop in `webrtc.go` and the voicemail greeting recorder in `voicemail.go`). Because the channel was never closed, those ranges blocked forever after each call ended, leaking one goroutine and one retained `*Pipeline` per completed call. On a Pi Zero 2 W with long uptime and many calls this is unbounded goroutine and heap growth.

`captureLoop` is the only sender on `outPCM` and is registered in `p.wg`, so `p.wg.Wait()` guarantees it has returned before the close: no send-on-closed-channel risk. `Stop()` cannot be called twice on the same instance (both hangup and greeting-finalize snapshot-and-nil `d.pipeline` under `d.mu`, so only one caller ever owns a given `*Pipeline`), so no double-close guard is needed.

This also fixes the audio self-test drain loop in `main.go`, which used `select { case <-OutFrames(): ...; default: goto drained }`. Once the channel is closed, a receive is always ready, so `default` would never fire and the loop would busy-spin. It now uses `for range` to drain buffered frames and exit on close.

## Test plan

- `go build ./...` and `go vet ./...` clean for the digitsd module.
- `go test ./internal/audio/...` passes.

---

### DIFF
--- a/pi/digitsd/cmd/digitsd/main.go
+++ b/pi/digitsd/cmd/digitsd/main.go
@@ -1728,16 +1728,12 @@ func main() {
 			}
 		}
 		pip.Stop()
-		// Drain any frames buffered between deadline and Stop
-		for {
-			select {
-			case frame := <-pip.OutFrames():
-				recorded = append(recorded, frame...)
-			default:
-				goto drained
-			}
+		// Drain any frames buffered before Stop closed the channel. Ranging
+		// over the now-closed OutFrames yields the remaining buffered frames
+		// and then returns.
+		for frame := range pip.OutFrames() {
+			recorded = append(recorded, frame...)
 		}
-	drained:
 
 		var maxAmp int32
 		for _, s := range recorded {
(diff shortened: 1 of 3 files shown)


justinlindh:

---

justinlindh:
Added the suggested regression test in 3fc3e8e: `TestPipeline_StopClosesOutFrames` builds a pipeline without `Start()` (no ALSA needed), buffers two frames, calls `Stop()`, and asserts a `for range OutFrames()` drains both and terminates, with a 1s timeout so a regression fails loudly instead of hanging. Verified it times out and fails when the `close(p.outPCM)` is removed, and passes with it.

---

PR #811, opened 2026-07-02
fix(digitsd): set write deadline on signaling websocket sends

justinlindh:
## What

Adds a write deadline to signaling websocket sends so a half-open TCP connection can no longer freeze the daemon.

## Why

Signaling reads are bounded (45s read deadline via the ping handler) and the pong control write uses a 10s deadline, but ordinary application sends (`Send` and the initial register write) never called `SetWriteDeadline`. On a half-open TCP connection (Wi-Fi drop, AP restart), once the socket send buffer fills, `WriteMessage` blocks until the kernel TCP retransmit timeout, which is many minutes. The sole consumer `handleSignal` runs inline on the daemon's main select loop, so a stuck write freezes hook and keypad processing: the phone cannot dial or even hang up until TCP gives up.

Each application write now sets a `writeTimeout` (10s) deadline first, reusing the constant the pong write already used. All writes are serialized behind the existing `c.mu`, and each new `SetWriteDeadline` is placed inside the same locked region as its paired write, so no data race is introduced. `WriteControl` (pong) is unchanged: gorilla documents it as safe alongside other writers and it carries its own per-call deadline.

## Test plan

- `go test ./internal/signal/...` passes.
- `go vet ./internal/signal/...` and full digitsd module build clean.

---

### DIFF
--- a/pi/digitsd/internal/signal/client.go
+++ b/pi/digitsd/internal/signal/client.go
@@ -15,6 +15,12 @@ const (
 	// before considering the connection dead. Must be greater than the
 	// server's ping interval (30s).
 	pingTimeout = 45 * time.Second
+
+	// writeTimeout bounds how long any write to the connection may block.
+	// Without it, a half-open TCP connection (WiFi drop / AP restart) with a
+	// full send buffer would block WriteMessage until the kernel TCP
+	// retransmit timeout (minutes), freezing the daemon's main select loop.
+	writeTimeout = 10 * time.Second
 )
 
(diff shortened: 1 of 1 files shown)


justinlindh:

---

justinlindh:
Agreed, `Close()` was the same failure class on the same inline main-loop path (pairing-code expiry and shutdown). Fixed in 330c298 by sending the close frame via `WriteControl` with a `writeTimeout` deadline, the gorilla-recommended way and matching the pong write. Builds, vets, and `internal/signal` tests pass.

---

PR #812, opened 2026-07-02
fix(digitsd): stop routing the DIAL:RESET:OK ack to onDial

justinlindh:
## What

Classifies the firmware's `DIAL:RESET:OK` ack as a fire-and-forget ack so it is no longer forwarded to the controller as a dialed number.

## Why

`DIAL:RESET:OK` is the ack the Pico sends in reply to the fire-and-forget `DIAL:RESET` command. But `isUnsolicitedEvent` returns true for any `DIAL:` prefix, and that check runs before the command-response and known-response paths, so the ack was delivered to the events channel. Downstream, `main.go` routes any `DIAL:` event to `onDial`, so the ack arrived as if the user had dialed the number `RESET:OK`, which runs the contact-filter/reject path or attempts a bogus call. The firmware only ever emits two `DIAL:` lines: `DIAL:RESET:OK` (this ack) and `DIAL:<digits>` (a real dialed number), so the fix is unambiguous.

`DIAL:RESET` is always sent via `SendFire` (no waiting consumer), so its ack should simply be dropped. The change excludes `DIAL:RESET:OK` from `isUnsolicitedEvent`, adds it to `isFireAndForgetAck` (dropped when no command is pending), and removes it from `isKnownResponsePrefix` so there is a single source of truth. A synchronous `DIAL:RESET` would still work: the pending-command path delivers any line to a waiter before the ack classifier runs.

## Test plan

- `go test ./internal/phone/...` passes.
- Extended `TestIsUnsolicitedEvent` to assert `DIAL:RESET:OK` is a command response, not unsolicited.
- Extended `TestIsFireAndForgetAck` to assert `DIAL:RESET:OK` is a fire-and-forget ack. Both assertions fail against the old classifier and pass with the fix.

---

### DIFF
--- a/pi/digitsd/internal/phone/serial.go
+++ b/pi/digitsd/internal/phone/serial.go
@@ -287,6 +287,12 @@ func isUnsolicitedEvent(line string) bool {
 		return true
 	case strings.HasPrefix(line, "KEY:"):
 		return true
+	case line == "DIAL:RESET:OK":
+		// Ack for the fire-and-forget DIAL:RESET command, not a dialed
+		// number. It shares the DIAL: prefix but must not be forwarded as an
+		// unsolicited event, or it reaches onDial as if the user dialed
+		// "RESET:OK". Handled as a fire-and-forget ack below instead.
+		return false
 	case strings.HasPrefix(line, "DIAL:"):
 		return true
 	case strings.HasPrefix(line, "FSM:"):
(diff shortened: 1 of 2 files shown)


justinlindh:

---

justinlindh:
Thanks for the review. Skipping the suggested `readLoop`-level integration test: it would require a fake `serial.Port` (no such harness exists here), and the two classifier assertions plus the fixed dispatch order in `readLoop` (unsolicited, then pending-command, then fire-and-forget-ack) already determine that a no-pending `DIAL:RESET:OK` is dropped before the known-response event delivery. The added cost of standing up port mocking is not worth it for behavior the unit tests already pin. The firmware-contract note (a future `DIAL:<sub>:OK` ack) is real but YAGNI until such a message exists.

---

